PR #3, opened 2019-07-11
Initial upload of hive protocol specification.

nagydani:
This is just to kick off the SWIP process and assign a SWIP number. Please DO NOT merge, until reade.


FantasticoFox:
I think the name should be "Hive Protocol". @nagydani

PR #7, opened 2019-07-11
Adaptive Node Protocol

FantasticoFox:
Starting SWIP process for the Adaptive Node Protocol

### DIFF
--- a/SWIPs/swip-adaptive-node-protocol.md
+++ b/SWIPs/swip-adaptive-node-protocol.md
@@ -1,0 +1,96 @@
+---
+SWIP: <to be assigned>
+title: Adaptive nodes capabilities protocol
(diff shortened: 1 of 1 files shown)


SWIPs/swip-adaptive-node-protocol.md
```diff
@@ -0,0 +1,96 @@
+---
+SWIP: <to be assigned>
+title: Adaptive nodes capabilities protocol
+author: Louis Holbrook @nolash <rhabdo@ethswarm.org>, Tim Bansemer (editorial) @FantasticoFox <tim@ethswarm.org>
+status: Draft
+type: Track Networking
+created: 2019-07-11
+---
+
+## Simple Summary
```
nolash: I would suggest that we merge `Simple Summary` and `Abstract` into one heading. As long as a sufficiently descriptive title is chosen, I don't see a need for both.

PR #11, opened 2019-08-01
Starting SWIP for Adaptive Node Protocol (Louis Swip)

zelig:
recreated from https://github.com/ethersphere/SWIPs/pull/7

PR #12, opened 2019-08-02
Adaptive node protocol

nolash:
obsoletes https://github.com/ethersphere/SWIPs/pull/11

### DIFF
--- a/SWIPs/swip-0.md
+++ b/SWIPs/swip-0.md
@@ -8,9 +8,9 @@ created: 2019-07-08
 updated: 2019-07-09
 ---
 
-## What is an SWIP?
+## What is a SWIP?
 
-SWIP stands for Swarm Improvement Proposal. An SWIP is a design document providing information to the Swarm community, or describing a new feature for Swarm or its processes or environment. The SWIP should provide a concise technical specification of the feature and a rationale for the feature. The SWIP author is responsible for building consensus within the community and documenting dissenting opinions.
+SWIP stands for Swarm Improvement Proposal. A SWIP is a design document providing information to the Swarm community, or describing a new feature for Swarm or its processes or environment. The SWIP should provide a concise technical specification of the feature and a rationale for the feature. The SWIP author is responsible for building consensus within the community and documenting dissenting opinions.
 
 ## SWIP Rationale
 
@@ -29,7 +29,7 @@ There are three types of SWIP:
 - A **Meta SWIP** describes a process surrounding Swarm or proposes a change to (or an event in) a process. Process SWIPs are like Standards Track SWIPs but apply to areas other than the Swarm protocol itself. They may propose an implementation, but not to Swarm's codebase; they often require community consensus; unlike Informational SWIPs, they are more than recommendations, and users are typically not free to ignore them. Examples include procedures, guidelines, changes to the decision-making process, and changes to the tools or environment used in Swarm development. Any meta-SWIP is also considered a Process SWIP.
 - An **Informational SWIP** describes an Swarm design issue, or provides general guidelines or information to the Swarm community, but does not propose a new feature. Informational SWIPs do not necessarily represent Swarm community consensus or a recommendation, so users and implementers are free to ignore Informational SWIPs or follow their advice.
 
-It is highly recommended that a single SWIP contain a single key proposal or new idea. The more focused the SWIP, the more successful it tends to be. A change to one client doesn't require an SWIP; a change that affects multiple clients, or defines a standard for multiple apps to use, does.
+It is highly recommended that a single SWIP contain a single key proposal or new idea. The more focused the SWIP, the more successful it tends to be. A change to one client doesn't require a SWIP; a change that affects multiple clients, or defines a standard for multiple apps to use, does.
 
 A SWIP must meet certain minimum criteria. It must be a clear and complete description of the proposed enhancement. The enhancement must represent a net improvement. The proposed implementation, if applicable, must be solid and must not complicate the protocol unduly.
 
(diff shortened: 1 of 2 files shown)


nagydani:
@nolash , I think, the latest commit, namely
https://github.com/ethersphere/SWIPs/pull/12/commits/1aafb847aa2758be2ef8a976587da81fa8f8fbbd
is unnecessary. It is contingent on your proposed changes to SWIPs in general, which I think are somewhat controversial. Also, I am wondering why you decided to use a different email address?
Would you be willing to revert that last commit?

nolash:
@nagydani I'm not sure what you mean; the latest commit is https://github.com/ethersphere/SWIPs/pull/12/commits/bd568752403056c1097b0fdafb738fc31350f38e

I don't use that email address, it's only for evil google aliasing.

nagydani:
Okay, then it's not the latest, but it is the commit that I linked, which makes it compliant with the changes in the template that are still pending.

SWIPs/swip-0.md
```diff
@@ -8,9 +8,9 @@ created: 2019-07-08
 updated: 2019-07-09
```
decanus: This seems like it should be in another PR, can we move it please?

SWIPs/swip-adaptive-node-protocol.md
```diff
@@ -0,0 +1,131 @@
...
+Capabilities {
+	Caps []Capability
+}
+```
+
+_Note that the RLP serialization of this structure in the golang implementation adds an extra list entry for every struct encapsulation. This "list of capabilities" will in reality be a "list of list of capabilities," where the former list only has one element._
+
+Furthermore, the bzz handshake message itself should also embed this message instead of the `Light` field that is currently there. Incidentally, in the golang implementation, the `BzzAddr` object transmitted in the bzz handshake is also (part of) the peer record stored by the kademlia peer database. A practical approach is therefore to embed the `Capabilities` information in the same structure. `BzzAddr` thus becomes:
+
+type BzzAddr struct {
```
diegomasini: Missing block code delimiters (```)

SWIPs/swip-adaptive-node-protocol.md
```diff
@@ -0,0 +1,131 @@
...
+	OAddr        []byte
+	UAddr        []byte
+	Capabilities *Capabilities
+}
```
diegomasini: Missing block code delimiters (```)

SWIPs/swip-0.md
oskarth: +1

SWIPs/swip-adaptive-node-protocol.md
```diff
@@ -0,0 +1,131 @@
...
+title: Adaptive nodes capabilities protocol
+author: Louis Holbrook <dev@holbrook.no> (https://holbrook.no), Tim Bansemer (editorial) @FantasticoFox <tim@ethswarm.org>
+status: Draft
+type: Track Networking
+created: 2019-07-11
+---
+
+## Abstract
+
+Adaptive node operation means a node should be able to change the capabilities it has according to changes in available resources.
```
oskarth: I'd specify, "available resources, such as bandwidth, connectivity ,...". This helps guide the reader what it's about

SWIPs/swip-adaptive-node-protocol.md
```diff
@@ -0,0 +1,131 @@
...
+
+## Motivation
+
+The peers of the node may need to adjust their behavior towards the node when these capabilities change. Therefore, a protocol is required to inform the peers of changes in capabilities.
+
+## Specification
+
+Upon initial connection, capabilities can be transmitted already within the handshake message.
```
oskarth: Perhaps this can point to spec that outline the handshake message?

SWIPs/swip-adaptive-node-protocol.md
```diff
@@ -0,0 +1,131 @@
...
+
+## Abstract
+
+Adaptive node operation means a node should be able to change the capabilities it has according to changes in available resources.
+
+## Motivation
+
+The peers of the node may need to adjust their behavior towards the node when these capabilities change. Therefore, a protocol is required to inform the peers of changes in capabilities.
+
+## Specification
```
oskarth: The first part reads more like an introduction, perhaps it can be under such a header? Then more clear definitions under Definitions header, Flow, Wire/data layout etc

SWIPs/swip-adaptive-node-protocol.md
```diff
@@ -0,0 +1,131 @@
...
+
+Upon initial connection, capabilities can be transmitted already within the handshake message.
+
+Capabilities are combinational. Some combination may make more sense than other. Interpreting which combinations are feasible and which are not should be handled by respective nodes, and not be hardwired in the protocol.
+
+For this reason, a bit vector seems the most logical choice. The bit vector can represent the adaptive node operation state within the node, and be embedded into the protocol message as-is.
```
oskarth: "seems" is a bit a of a weasel word in a spec. Better to just state what it does

SWIPs/swip-adaptive-node-protocol.md
```diff
@@ -0,0 +1,131 @@
...
+Since the capabilities may have to be set from different system modules, the network module should provide an API for registering them. The following structure is proposed:
```
oskarth: why proposed? as opposed to "this is the structure"

SWIPs/swip-adaptive-node-protocol.md
```diff
@@ -0,0 +1,131 @@
...
+
+* the first byte is the identifier of the module (0x00 is bzz, and built in)
```
oskarth: this might be clearer if laid out as ascii data layout or so, then explain further

e.g. see rfc style specs

SWIPs/swip-adaptive-node-protocol.md
```diff
@@ -0,0 +1,131 @@
...
+* the second byte is the length `n` of the bit vector containing the flags
+* the last `n` bytes is the bit vector containing the flags
+
+For the individually capability flags we will initially be choosing from, we can see an example of two distinct byte pairs representing tematic categories:
```
oskarth: individual

this sentence can be more concise and to the point

SWIPs/swip-adaptive-node-protocol.md
```diff
@@ -0,0 +1,131 @@
...
+
+```
+BZZ/SYNC: 			// (registered by default)
```
oskarth: point to relevant specs like bzz/sync and pss (right below this section)

SWIPs/swip-adaptive-node-protocol.md
```diff
@@ -0,0 +1,131 @@
...
+- retrieve		0x00020001
```
oskarth: all these capabilities should be elaborated on right after in terms of what they mean

SWIPs/swip-adaptive-node-protocol.md
```diff
@@ -0,0 +1,131 @@
...
+
+Furthermore, the bzz handshake message itself should also embed this message instead of the `Light` field that is currently there. Incidentally, in the golang implementation, the `BzzAddr` object transmitted in the bzz handshake is also (part of) the peer record stored by the kademlia peer database. A practical approach is therefore to embed the `Capabilities` information in the same structure. `BzzAddr` thus becomes:
+
+type BzzAddr struct {
+	OAddr        []byte
+	UAddr        []byte
+	Capabilities *Capabilities
+}
+
+Consequently the `LightNode` flag is removed from the `HandshakeMsg` message.
```
oskarth: this is hard to parse, "is removed"? the spec should state what is the case, if things are changing or there's a diff from previous versions, I suggest we call that out in a separate section

SWIPs/swip-adaptive-node-protocol.md
```diff
@@ -0,0 +1,131 @@
...
+
+```
+HandshakeMsg {
+	Version		uint64
+	NetworkID	uint64
+	Addr		*BzzAddr
+}
+```
+
+For simplicity, arrays of bools are used to represent the flags in interfacing code. In the JSON-RPC API a `Capability` is represented in the following manner:
```
oskarth: implementation detail, perhaps add to implementation suggestion section to keep spec more concise?

SWIPs/swip-adaptive-node-protocol.md
```diff
@@ -0,0 +1,131 @@
...
+For example, a `Capability` with id 42 and three flags with values `true, false, true` this becomes:
+
+```
+{
+	"Id": 42,
+	"Cap": [true, false, true],
+}
+```
+
+The network module API can merely consist of:
```
oskarth: again, "can" - I want to know what it does

PR #16, opened 2019-08-23
SPEC: Basis datatype representations for protocols in node implementer specs

nolash:
This is the SWIP markdown rendering of the [data.latex](https://github.com/ethersphere/node-implementer-spec/blob/master/src/data.latex) document from the node-implementer-spec.

It defines the format in which datatypes will be described, and defines basic data type components required to build the respective protocol descriptions for Swarm.

----

### DIFF
--- a/SWIPs/swip-data-types-descriptions.md
+++ b/SWIPs/swip-data-types-descriptions.md
@@ -7,13 +7,13 @@ type: General
 created: 2019-08-02
 ---
 
-## abstract
+## Abstract
 
-This SWIP defines pseudo data types for use in Swarm protocol specifications.
+This SWIP defines pseudo data types for use in Swarm protocol specifications, and their serialization to RLP.
 
-## motivation
+## Motivation
 
-To describe protocols in our specifications, we are in need of a strongly defined way of describing data types and their serializations to RLP.
+To describe protocols in our specifications, we are in need of a strongly defined way of describing data types and their serializations to RLP. RLP is the relevant format in this case, as it is the serialization format currently being used in devp2p, and Swarm is currently based on devp2p.
 
 RLP does not have "types" as such, just self-describing data units whose length is serially interpreted. Hence it is for example not possible to define a number value range merely in terms of RLP constructs.
 
@@ -64,7 +64,7 @@ All other data is encoded as RLP lists.
 
 The outermost element of a serialization is always a LIST item [^1]
 
-## example
+### example
 
 A message structure with only one field of `3OCTET` data will in
 *reality* be `LIST(3OCTET)`. If the data is `0x666f6f` it will serialize
@@ -75,3 +75,19 @@ LIST:   0xc4 (list, 4 bytes long)
 3OCTET: 0x83 (string, 3 bytes long)
 data:   0x66 0x6f 0x6f
 ```
+
+## Backwards Compatibility
+
+There are no compatibility issues.
+
+## Test Cases
+
+There are no relevant test cases.
+
+## Implementation
+
+Any future SWIP defining Swarm protocols.
+
+## Copyright
+
+Copyright and related rights waived via [CC0](https://creativecommons.org/publicdomain/zero/1.0/).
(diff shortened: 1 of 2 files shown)


SWIPs/swip-data-types-descriptions.md
```diff
@@ -0,0 +1,66 @@
...
+### data representations
+
+This document uses the `ABNF` convention for defining data types, as
+described in the IETF RFC 5234. 
+
+#### primitives
+
+The basic identifiers are defined as follows:
+
+    UINT64MAX   = 18446744073709551615
```
oskarth: table format

SWIPs/swip-data-types-descriptions.md
```diff
@@ -0,0 +1,66 @@
...
+Copyright and related rights waived via CC0
+
+## Specification
+
+### data representations
+
+This document uses the `ABNF` convention for defining data types, as
+described in the IETF RFC 5234. 
+
+#### primitives
```
oskarth: capitalize

SWIPs/swip-data-types-descriptions.md
```diff
@@ -0,0 +1,66 @@
...
+    
+    BOOL        = BIT
+    
+    TIMESTAMP   = UINT32
+
+#### network specific types
+
+#### rule extension for list definitions
+
+The construct `LIST` explicitly defines a list of elements. Elements may
```
oskarth: Can we write this in some more precise formatting form? A la bnf or so

SWIPs/swip-data-types-descriptions.md
nolash: I've changed it to table, but I must admit I prefer the classic ABNF definition syntax. I also thought we agreed on keeping this representation for the ABNF, but instead tables for other definition lists.

One reason to avoid tables in this case is the superfluous header column that is required for md tables.

SWIPs/swip-data-types-descriptions.md
nolash: Is this ok?

SWIPs/swip-data-types-descriptions.md
```diff
@@ -0,0 +1,70 @@
...
+
+## Copyright
+
+Copyright and related rights waived via CC0
+
+## Specification
+
+### Data representations
+
+This document uses the `ABNF` convention for defining data types, as
```
oskarth: link to rfc

SWIPs/swip-data-types-descriptions.md
```diff
@@ -0,0 +1,70 @@
...
+
+The basic identifiers are defined as follows:
+
+| id  | def |
+| --- | :--- |
+|    UINT64MAX  | 18446744073709551615 |
+|    UINT32MAX  | 4294967295 |
+|    UINT16MAX  | 65535 |
+|    UINT8MAX   | 255 |
+|    UINT64     | %d0-UINT64MAX |
```
oskarth: not sure, but perhaps it is good to elaborate in this since it won't be obvious to everone reading spec what "%d0" is

SWIPs/swip-data-types-descriptions.md
```diff
@@ -0,0 +1,70 @@
...
+This document uses the `ABNF` convention for defining data types, as
+described in the IETF RFC 5234. 
+
+#### Primitives
+
+The basic identifiers are defined as follows:
+
+| id  | def |
+| --- | :--- |
+|    UINT64MAX  | 18446744073709551615 |
```
oskarth: is this a "round" number? if so, might be better to express as such, e.g. `2^32 -1` or whatever

SWIPs/swip-data-types-descriptions.md
oskarth: hm alright, forgot :sweat: different schools perhaps?

>  superfluous header column

usually the header column is meaningful, what's an example of where it isn't?

cc @decanus

nolash:
@nagydani TIMESTAMP it is the common bitlength for timestamp types that I know of at least on *nix platforms.

PR #17, opened 2019-08-23
SPEC: Swarm overlay address definition

nolash:
This part of the spec defines the Swarm Overlay Address in the highest level terms.

The devp2p specific information is omitted, but the current state of the description of it can be found in the appendices in the reference document (see below).

----


SWIPs/swip-overlay-address.md
```diff
@@ -0,0 +1,67 @@
...
+## Specification
+
+### Base Hash
+
+Swarm uses `KECCAK256` as its `Base Hash` algorithm. `KECCAK256` has a
+digest size of 32 bytes.
+
+### Swarm Overlay Address
+
+Swarm addresses all content and all nodes using algorithms designed
```
oskarth: Here it seems like it is used for:
- content
- nodes

However, here only node addresses are mentioned. For continuity, it might be good to elaborate on content too (even if if this is just a pointer to a different spec)

SWIPs/swip-overlay-address.md
```diff
@@ -0,0 +1,67 @@
...
+around its `Base Hash`.
```
oskarth: "around" seems vague, can we be more precise?

SWIPs/swip-overlay-address.md
```diff
@@ -0,0 +1,67 @@
...
+Formally we define the `Swarm Overlay Address` thus:
+
+    OVERLAYADDRESS = 32*(%x00-ff)
+
+#### Swarm Address Pair
+
+To enable peers to locate the a node on the network, the aforementioned
+`Swarm Overlay Address` is paired with an Underlay Address.
+
+The Underlay Address is a string representation of the node’s network
```
oskarth: I sometimes find the formatting a bit a bit inconsistent. E.g you have

`Swarm Overlay Address`
Underlay Address
   UNDERLAYADDRESS

Perhaps we can use only one format for Proper Nouns and stick to it? Should help with visual parsing and searching

nolash: I think the ABNF notation gets unclear if we use spaces in the identifiers. 

And I find the descriptive text gets ugly if we use identifiers.

I agree that `Swarm Overlay Address` and "Underlay Address" in the same form.

Is this possible middle ground:

* One form for the ABNF defs
* One form for glossary definitions in the main document.

oskarth: lets try it

oskarth: ^ @decanus

nolash:
@nagydani regarding base hash/keccak https://github.com/ethersphere/SWIPs/pull/31 

how should the test case look here, exactly? Just some input and expected output?

PR #18, opened 2019-08-23
Add capabilities data type and role description

nolash:
Adds a general description of the role of capabilities, and the high level data representation of them as they are communicated between nodes.

This description **SHOULD** reflect the current implementation on the master branch, aswell as the golang reference implementation SWIP on the same topic in https://github.com/ethersphere/SWIPs/pull/12 

----


nolash:
@diegomasini this is already done no? https://github.com/ethersphere/SWIPs/pull/18/files#diff-afecfc6fadb78338bd7edf5826f4ae29R2

PR #20, opened 2019-08-30
Swip honey to money

Eknir:
This SWIP proposes an oracle to resolve an amount of honey to Wei.


Eknir:
I assigned number `20`. Fixing the JSON errors is difficult, as the abstract example will never be appropriate JSON (e.g. `...` or <numberN> is not allowed).

I fixed it by not telling anymore to MD that the abstract example is JSON, whereas I still tell this in the implemented example.

SWIPs/swip-20-honey_to_money.md
```diff
@@ -0,0 +1,103 @@
...
+Status: Draft
+Type: Standards track
+Category: Core
+Created: 30-08-2019
+---
+<!--You can leave these HTML comments in your merged SWIP and delete the visible duplicate text guides, they will not appear and may be helpful to refer to if you edit it again. This is the suggested template for new SWIPs. Note that a SWIP number will be assigned by an editor. When opening a pull request to submit your SWIP, please use an abbreviated title in the filename, `SWIP-draft_title_abbrev.md`. The title should be 44 characters or less.-->
+ 
+## Simple summary
+<!--"If you can't explain it simply, you don't understand it well enough." Provide a simplified and layman-accessible explanation of the SWIP.-->
+Swarm needs a mechanism to set prices which makes sense relative to other (distributed storage) services in the market. As the absolute price of using Swarm is defined by the cost of honey (Swarm's internal accounting unit), this SWIP proposes a minimal interface to enable upgrading honey prices, network-wide and for all nodes. The specifics on what an efficient price should reflect and how the update process looks like is *not* part of this SWIP.
```
nolash: Please consider replacing "upgrading" with "updating."

SWIPs/swip-20-honey_to_money.md
```diff
@@ -0,0 +1,103 @@
...
+
+## Abstract
+<!--A short (~200 word) description of the technical issue being addressed.-->
+It is expected that the price of using Swarm will change over time to reflect the changing economic landscape as well as to facilitate experimentation. Such price updates must happen atomically to prevent accounting unbalances from appearing. To lay the groundwork for enabling such price changes, this SWIP proposes a minimal interface for a *price-oracle*. Furthermore, this SWIP includes a proposal for implementing an *on-chain* price oracle, governed by Swarm developers and stakeholders to allow for experimentation.
+ 
+This SWIP is part of a series of SWIPs (but can be implemented on its own). To see the full picture, please refer to [swip-message_to_honey](https://github.com/Eknir/SWIPs/blob/SWIP-message_to_honey/SWIPs/swip-message_to_honey.md), [swip-multiple_payment_processing_support](https://github.com/Eknir/SWIPs/blob/SWIP-multiple_payment_processing_support/SWIPs/swip-multiple_payment_processing_support.md) and the diagram below:
+![SWIP_Diagrams.png](./../assets/swip-honey_to_money/SWIP_Diagrams.png)
```
nolash: To the diagram:

* Do lines terminating outside nodes have any significant meaning as opposed to lines intersecting the nodes?

* Why is there a line connecting process message, disconnect threshold reached condition (no), update balance. The "no" seems to go in both directions, and terminate nowhere.

SWIPs/swip-20-honey_to_money.md
```diff
@@ -0,0 +1,103 @@
...
+ 
+## Motivation
+<!--The motivation is critical for SWIPs that want to change the Swarm protocol. It should clearly explain why the existing protocol specification is inadequate to address the problem that the SWIP solves. SWIP submissions without sufficient motivation may be rejected outright.-->
+Nodes keep track of balances relative to each other using an internal reference unit: honey. Since honey is not a currency in which nodes can settle their balances with each other—it is merely a unit of account—there is a requirement to facilitate converting honey to a currency. Furthermore, it is desirable that this conversion rate can fluctuate to reflect the changing economic landscape and facilitate experimentation.
```
nolash: > Since honey is not a currency

Please consider defining "honey" more clearly, and earlier in the document. In general I find it clearer if term definition comes before usage.

SWIPs/swip-20-honey_to_money.md
```diff
@@ -0,0 +1,103 @@
...
+
+This update process should be atomic: either all nodes upgrade or none of them do. The current implementation hardcodes honey prices in the source code. This violates the atomicity requirement as there is no guarantee that all Swarm nodes are on the same release. By externalizing the price to an oracle, Swarm completely decouples its core functions from price discovery. By building support for price oracles, it will be possible to coordinate price changes across the network and set the stage for future experimentation with various mechanisms for setting prices, from fixed prices to DAOs or to bonding curves.
```
nolash: Update and upgrade is not the same term. Please clarify.

nolash: > By externalizing the price to an oracle, Swarm completely decouples its core functions from price discovery.

The choice is not binary between and oracle and having a hardcoded number in the implementation. I don't see a coherent argument here presented for why an oracle is the only way to reach consensus on per-message cost in terms of unit of account.

SWIPs/swip-20-honey_to_money.md
```diff
@@ -0,0 +1,103 @@
...
+
+## Specification
+<!--The technical specification should describe the syntax and semantics of any new feature. The specification should be detailed enough to allow competing, interoperable implementations for the current Swarm platform and future client implementations.-->
+* The `honeyToMoneyOracle` (being an instance which runs on the Swarm node) fetches its data from one specific endpoint (the external oracle). This endpoint is included as a part of the `honeyToMoneyOracle` implementation in the Swarm source-code, but can be adjusted by the node if so desired (e.g. via config and/or flag).
+* If two nodes query the external oracle at the same time, the external oracle gives the same answer.
```
nolash: "at the same time" is a weak definition. Please improve.

SWIPs/swip-20-honey_to_money.md
```diff
@@ -0,0 +1,103 @@
...
+* A `honeyToMoneyOracle` returns the most up-to-date `honeyPrice` to other components of the Swarm.
+* Internally, the `honeyPrice` is decided by the `honeyToMoneyOracle` by building a strucure like the `honeyPrice` object. `honeyPrice` specifies a `Time To Live (TTL)` in seconds and a `price` object. The `price` object maps all `validFrom` entries to their respective `price`. `validFrom` is a UNIX timestamp and `price` is the price of one honey unit, denominated in `currency`. It is up to the implementation of the `honeyToMoneyOracle` to decide on the actual structure to use. To summarize, the oracle may return:
```
nolash: If the structure is implementation dependent, why is not instead an API described here?

SWIPs/swip-20-honey_to_money.md
```diff
@@ -0,0 +1,103 @@
...
+### General rationale for choosing oracles
+There exists a need for the price of Swarm to change over time. However, due to the nature of Swarm, there is a need for this price to be decided network-wide, instead of peer-to-peer. Specifically:
+- In order for a node to have Kademlia connectivity, it must be connected to all peers in his most proximate bin and *at least* one peer for each proximity order up to the saturation depth (see [architecture docs](https://swarm-guide.readthedocs.io/en/latest/architecture.html#overlay-network)). For SWAP to work, nodes must have the certainty that they will be able to reach an agreement on the price when they have to settle their balance. Hence, in the situation that nodes agree on the price peer-to-peer, this price *must* be decided upon the point when nodes connect. As it might happen that they don't come to an agreement on price and hence don't connect, Kademlia connectivity cannot be guaranteed when prices are to be decided peer-to-peer—in particular because of the requirement of connectivity to all nodes in the most proximate bin.
```
nolash: Please note that the term "saturation depth" is not well-defined in Swarm. It is never quite clear whether it refers to the depth used for routing decisions/peer retrievals from kademlia (where what you describe is that case), or whether it refers to the connectivity driver depth with requires _two_ connected nodes in a bin to be satisfied. I am currently advocating using the terms `Neighborhood Depth` and `Saturation Depth` to disambiguate them.

nolash: SWAP here is mentioned without definition. Will it ultimately be part of the same document, or should it have a reference link?

SWIPs/swip-20-honey_to_money.md
```diff
@@ -0,0 +1,103 @@
...
+- Zero marginal costs of serving data with perfect competition means that when nodes would engage in peer-to-peer price discovery, the price will tend to zero and a sub-optimal equilibrium of near-zero prices will be reached.
```
nolash: Hayek and Friedman would roll in their graves. 

This is a _very_ bold claim, and I would like to see this backed up with theory.

SWIPs/swip-20-honey_to_money.md
```diff
@@ -0,0 +1,103 @@
...
+- The high-quantity of transactions implies that nodes cannot engage in price-discovery for every byte of data transacted, meaning they have to agree *beforehand* on how they will come to a price *at a later time*.
+- Nodes are most of the time just conduits for data and payments, meaning that most of the time data requests (and thus also payments) will take place because nodes want to send data to a peer who requested this data from them.
```
nolash: I don't understand why this is an argument for oracles. P2p discovery agreeing currency cost per unit for the next `N` units (or next `T` seconds) is just as possible.

SWIPs/swip-20-honey_to_money.md
nolash: It would not hurt to split this bullet up into subsections.

SWIPs/swip-20-honey_to_money.md
```diff
@@ -0,0 +1,103 @@
...
+
+If there appears to be a problem with such imbalances, we see a possible solution: as nobody benefits from price imbalances, both the sender and the receiver of a payment are incentivized to apply a price which is not per their clock, but per the clock of the counterparty when they expect the counterparty to query or have queried the oracle. To assist in this, a payload might be sent with the transaction, specifying the clock with which the oracle price was applied.
```
nolash: > as nobody benefits from price imbalances

Why is this the case?

If I was the providing node and benefited from higher prices, can't I just adopt the attitude that nodes simply top up their credit to afford consumption?

nolash: > both the sender and the receiver of a payment are incentivized to apply a price which is not per their clock, but per the clock of the counterparty when they expect the counterparty to query or have queried the oracle. 

I don't understand this sentence. If both consult each others' clocks instead of their own, how is the result different than both consulting their own clocks. What dpes the sum of the two "clocks" determine?

SWIPs/swip-20-honey_to_money.md
```diff
@@ -0,0 +1,103 @@
...
+
+#### Temporary hard forks
+An on-chain oracle *might* give a different answer to two nodes when queried at the same time if there exists a (temporary) fork in the blockchain. If a disbalance due to such forks becomes common and causes the disconnect threshold to be reached for a significant number of nodes, we may put up additional requirements. For example, a minimum of 12 blocks for confirmations.
```
nolash: disbalance -> imbalance

nolash: > For example, a minimum of 12 blocks for confirmations.

Or just use price at `B` blocks ago by default. Or was that what you meant here?

SWIPs/swip-20-honey_to_money.md
```diff
@@ -0,0 +1,103 @@
...
+* The Swarm `honeyToMoneyOracle` will reference an Ethereum address of a `HoneyToMoney` price oracle smart contract. This smart contract will fulfill the role of the external oracle.
+* The currency in which the oracle quotes its price is Wei.
+* The `honeyToMoneyContract` implements the `HoneyToMoney` interface (to be specified). 
+* The `HoneyToMoneyContract` will initially be governed in such a way that a subset of stakeholders can make decisions which are not high-impact (i.e. decide on small price changes) but the involvement of all stakeholders is required to make high-impact decisions (such as changing the implementation of the oracle, changing governance, deciding on big price changes, etc). This semi-governance structure is in place to allow experimentation with an update process of honey prices and is not a recommendation for a price update process when honey is quoted in a currency worth real money.
```
nolash: I am sure people would be very interested in how this governance scheme will take place. Is this a separate SWIP? Since you mention other what elements are separate SWIPs other places in this document, plesae consider whether it's worth a mention here, too.

SWIPs/swip-20-honey_to_money.md
Eknir: Done

SWIPs/swip-20-honey_to_money.md
Eknir: > Do lines terminating outside nodes have any significant meaning as opposed to lines intersecting the nodes?

What do you mean here?

> Why is there a line connecting process message, disconnect threshold reached condition (no), update balance. The "no" seems to go in both directions, and terminate nowhere.

Sorry for the confusion here. When I checked it just now, it also took me a while to figure out, meaning it was not clear. Process message is an end-state and can only be reached via `should it be accounted for, no`, `update balance` (debitor) and `disconnect threshold reached, no` (creditor).

`Disconnect threshold reached, no => process message` is the only allowed direction. I will have a look together with @vojtechsimetka how we can make this clear. Thanks.

> Please consider three separate diagrams under each other; it's too small to be legible in the markdown document, and the enlargement behind the first click is a frame with scroll.

We will have a look at how this can be improved.

SWIPs/swip-20-honey_to_money.md
Eknir: Update is the correct term. Updated ;)

SWIPs/swip-20-honey_to_money.md
Eknir: > disbalance -> imbalance

Changed.

> Or just use price at B blocks ago by default. Or was that what you meant here?

This is what is meant here, but I will rewrite it to something which is more in line with how you phrased it

SWIPs/swip-20-honey_to_money.md
Eknir: @nolash , I started working on an implementation of the simple governance scheme described here [see here](https://github.com/Eknir/msgOracle). AFAIK, there is no intention to write a SWIP for this yet, as the idea is to first get some experience. 

When we are going to run Swarm on a live network, there must be a clear idea how the price is going to be updated, what the process around this is (at that time) and how we envision updating this process to give more autonomy to the users of Swarm.

SWIPs/swip-20-honey_to_money.md
Eknir: Added a paragraph defining honey in the abstract section

SWIPs/swip-20-honey_to_money.md
Eknir: This is described in the rationale sector, but to be honest with you: I am myself also not fully convinced of this being the way-to-go. 
To be clear: this is not about a _per-message cost_, but it is about the price you pay for resetting the balance with your peer back to zero, given an imbalance. 

If you welcome the idea, I would be happy to discuss with you the current rationale behind chosing the oracle and you can either help me to sharpen this rationale, or bash it completely (in which case we have to come up with a different solution). 

FYI: I have asked a professor in micro-economics/auction theory/game design to help me brainstorming on exactly this topic, so I would welcome to speak about it with you as well.

SWIPs/swip-20-honey_to_money.md
Eknir: done

SWIPs/swip-20-honey_to_money.md
Eknir: > Why is this the case?

A price imbalance comes into existence if one node (A) applies a higher price than the other (B). Sure, if A is the net payer, B will benefit, but here it is meant that B is the net user, hence the price balance means that A get's less than expected, while B pays exactly what he thinks he should pay. This becomes a problem when this price balance becomes sufficiently big to lead to a disconnect and nobody benefits from disconnects.

I would rephrase it as: `as nobody benefits from price imbalances that lead to disconnects, ...`

Eknir: Please note that this section part is under the discussion sector. If we were to make an implementation based on this, we definitely have to clarify more, but let me give an example of what this *could* mean:

If there is indeed a delay between the time when a receiver of payment queries the oracle and when the sender does so and this delay causes disconnects, we could add an extra field which stipulates at which block the oracle was/will be queried with respect to this transaction.

The counterparty now knows the exact time a conversion happened and can query the blockchain at that time, to get the same conversion. 

Obviously, this query has to be somewhere in between bounds: if a sender of a payment tells he did the conversion 2 years ago, the other node will obviously not apply that conversion and just use the most recent conversion.

That was an example of the receiver applying the conversion "per the clock of the sender". A sender of payment can obviously also predict at what time a receiver will query the oracle, and can signal in the payload that he will defer querying the oracle up until that time. In this case, the sender queries the oracle per the clock of the receiver.

Again, these ideas are not worked out yet, but it might be a possible solution if it happens that imbalances due to a time delay between querying the oracle is a problem.

SWIPs/swip-20-honey_to_money.md
nolash: I'm sorry, but I feel to do justice to a deeper discussion on the topic requires a certain degree of immersion, and I don't think I have enough time to set aside for that. 

As an audience, however, I think what I feel is missing at a minimum some reference to theoreical sources to back up the claims that are being made in this SWIP.

SWIPs/swip-20-honey_to_money.md
vojtechsimetka: > Why is there a line connecting process message, disconnect threshold reached condition (no), update balance. The "no" seems to go in both directions, and terminate nowhere.

Nice catch! Should be good now

SWIPs/swip-20-honey_to_money.md
Eknir: Thank @nolash , this will be addressed.

PR #21, opened 2019-08-30
Swip message to honey

Eknir:
This SWIP proposes an on-chain oracle to resolve the price for any Swarm message type in honey


SWIPs/swip-X.md
```diff
@@ -0,0 +1,54 @@
+---
```
oskarth: leftover commit?

PR #22, opened 2019-08-30
Swip multiple payment processing support

diegomasini:
This SWIP proposes modifications to the Swarm client to incorporate support for payment modules. A payment module encapsulates all the payment logic for a specific payment technology (e.g. Lumino, Raiden, etc.) and allows the Swarm client to delegate payment processing to these modules. Swarm peers will negotiate which payment module they want to use to settle their debts for the services consumed.


diegomasini:
> Please assign number 22, this one is ready to receive Draft status.

Done!

SWIPs/swip-22-multiple_payment_processing_support.md
```diff
@@ -0,0 +1,464 @@
...
+        providers:
+            swap: 90
+        oracles:
+            ethOracleA: 90
+            ethOracleB: 10
+```
+
+When two nodes (A and B) establish a connection they independently execute the following payment method selection algorithm, using as input the configuration defined previoulsy:
+
+1. Based on the node configuration generate a set of triplets in the form ```[currency, provider, oracle]```. As an example let's take from the previous configuration the ```rif``` entry. From that entry we can build the triplets `[rif, lumino, rifOracleA]`, `[rif, lumino, rifOracleB]`, `[rif, lumino, rifOracleC]`,`[rif, raiden, rifOracleA]`, `[rif, raiden, rifOracleB]`, `[rif, raiden, rifOracleC]`, but triplets are generated for the other currencies as well.
```
oskarth: What are the requirements for nodes communicating across different implementations of Swarm? E.g. wire format, protobuf etc... is that out of scope of this document?

diegomasini: From my point of view that is out of the scope of this document.

SWIPs/swip-X.md
```diff
@@ -0,0 +1,54 @@
...
+
+## Backwards Compatibility
+<!--All SWIPs that introduce backwards incompatibilities must include a section describing these incompatibilities and their severity. The SWIP must explain how the author proposes to deal with these incompatibilities. SWIP submissions without a sufficient backwards compatibility treatise may be rejected outright.-->
+All SWIPs that introduce backwards incompatibilities must include a section describing these incompatibilities and their severity. The SWIP must explain how the author proposes to deal with these incompatibilities. SWIP submissions without a sufficient backwards compatibility treatise may be rejected outright.
+
+## Test Cases
+<!--Test cases for an implementation are mandatory for SWIPs that are affecting changes to data and message formats. Other SWIPs can choose to include links to test cases if applicable.-->
+Test cases for an implementation are mandatory for SWIPs that are affecting changes to data and message formats. Other SWIPs can choose to include links to test cases if applicable.
```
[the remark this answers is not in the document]
holisticode: On the other hand, I understand this SWIP to be just a conceptual approaching to the implementation of a multi-payment solution. But then, if we want to take the SWIP process seriously, and if I understand the process correctly, then we should create an additional SWIP with the actual implementation which would also include test cases? This then relates to the implementation chapter: "rough consensus and running code", so I would expect some words in related to test cases and implementation?

PR #24, opened 2019-09-05
SPEC: Kademlia

nolash:
**This is work in progress.**

The full evolving spec document can be viewed here:

* (latex source) https://github.com/ethersphere/node-implementer-spec
* (pdf) https://swarm-gateways.net/bzz:/dcad67ce1b19aa80d8bb2c1e5c2244c2e67f3f0f894cf863ed5522d0943e4c14/


nolash:
@nagydani Regarding your bullet points on kademlia properties; are any of these properties particular to the swarm implementation? If they are all common properties, and by this I mean common in the sense that the same guarantees are provided by the Kademlia Paper by Maymounkov/Mazieres, then I think it makes more sense to merely refer to that paper and say something like:

> The Swarm implemenation of Kademlia provides the usual guarantees of logarithmic proportions regarding connectivity and routing.

nolash:
> I also think that we use (or should use) a bit more complex Neigborhood definition. Let's discuss.

My understanding is that we _currently do not use_ a more complex definition. Can you please confirm to your knowledge whether or not this is the case?

The main purpose of this spec is to describe the actual state of implementation. If changes are to be made, we adjust the document when we have decided on those changes. Otherwise we are at risk of ever iteration and never getting anywhere.

That said; how do we proceed to discuss further development of the neighborhood definition?

SWIPs/swip-23-kademlia.md
```diff
@@ -0,0 +1,238 @@
...
+
+N/A
+
+## Copyright
+
+Copyright and related rights waived via CC0
+
+## Specification
+
+The Kademlia is both a database (DHT) of known and connected peers on
```
oskarth: I'd put DHT in a definitions table and write it out in full for people who don't know

SWIPs/swip-23-kademlia.md
```diff
@@ -0,0 +1,238 @@
...
+the network, as well as a component for routing and forwarding content
+on the network.
+
+### Records
+
+A peer record *ALWAYS* contains the `Address Pair` and the
```
oskarth: I'd write MUST to keep a concise shared vocabulary

SWIPs/swip-23-kademlia.md
```diff
@@ -0,0 +1,238 @@
...
+    BZZADDRESS = LIST(OVERLAYADDRESS UNDERLAYADDRESS CAPABILITIES)
+
+### Connectivity
+
+#### Address matching
+
+For two addresses to match each other, their byte values must be
+identical.
+
+For certain purposes in Swarm *partial addresses* are used. A partial
```
oskarth: So the other bits aren't padded out?

SWIPs/swip-23-kademlia.md
```diff
@@ -0,0 +1,238 @@
...
+that the partial address represents. A partial address matches a full
+address if all the bytes in the partial address are identical to the
+corresponding bytes in the full address.
+
+#### Distance and proximity
+
+In Swarm, differences between addresses are described in terms of
+*Distance* and *Proximity*.
+
+The *Distance* between two addresses is the value of the big-endian
```
oskarth: This seems like math in plain text, something like:

The *distance* between two addresses A and B is defined as follows:

`distance(a,b) = f(xor(a,b))`

where f (or whatever better name) is <explain>

SWIPs/swip-23-kademlia.md
```diff
@@ -0,0 +1,238 @@
...
+address is an address of less than 32 bytes. The bytes of a partial
+address always represent the most significant bytes of the full address
+that the partial address represents. A partial address matches a full
+address if all the bytes in the partial address are identical to the
+corresponding bytes in the full address.
+
+#### Distance and proximity
+
+In Swarm, differences between addresses are described in terms of
+*Distance* and *Proximity*.
```
oskarth: call out both as functions (which afaik they are?), like the distance function

then define clearly

SWIPs/swip-23-kademlia.md
```diff
@@ -0,0 +1,238 @@
...
+
+The *Distance* between two addresses is the value of the big-endian
+binary integer cast of the XOR of the two addresses.
+
+For example, the distance between the addresses 1010 and 1101 will be
+\(1011 ^ 0111 = 1100\), where the big-endian binary integer cast of
+\(1100\) is \(12\).
+
+Another measure of difference *Proximity*, which is the logarithmic
```
oskarth: Proximity Order format

SWIPs/swip-23-kademlia.md
```diff
@@ -0,0 +1,238 @@
...
+Order, which is index value of the most significant bit that differs
+between the two addresses.
+
+For example, an address \(0101\) will have a Proximity Order of \(0\)
+compared to \(1101\), a proximity Order of \(1\) compared to \(0011\),
+and a Proximity Order of \(3\) compared to \(0110\). 
+
+#### Proximity Bin
+
+A Proximity Bin is the table row in the kademlia that contains all peers
```
oskarth: "in the kademlia" a bit vague

SWIPs/swip-23-kademlia.md
oskarth: e.g. https://swarm-guide.readthedocs.io/en/latest/architecture.html#overlay-network

SWIPs/swip-23-kademlia.md
```diff
@@ -0,0 +1,238 @@
...
+
+The Depth of a node’s kademlia is found by the following method:
+
+1.  Find the two closest Proximity Bins that have at least one
+    *connected* peer.
+
+2.  Find the next farther Proximity Bin that has at least one
+    *connected* peer.
+
+3.  If found, the Depth is the numeric value of that bin plus one.
```
oskarth: math in plain text, belongs in definition/mathy/pseudocode

SWIPs/swip-23-kademlia.md
```diff
@@ -0,0 +1,238 @@
...
+A kademlia is healthy when:\[1\]
+
+  - All peers in the Neighborhood are connected
+
+  - At least one peer in every Proximity Bin *farther* than the Depth is
+    connected.
+
+### Content forwarding
+
+The rules for content forwarding apply to *all* nodes on the content
```
oskarth: emphasis and proper nouns have same formatting, I'd suggest decoupling these

SWIPs/swip-23-kademlia.md
```diff
@@ -0,0 +1,238 @@
...
+2.  Find the next farther Proximity Bin that has at least one
+    *connected* peer.
+
+3.  If found, the Depth is the numeric value of that bin plus one.
+
+4.  If not found, the depth is \(0\).
+
+#### Health
+
+A kademlia is healthy when:\[1\]
```
oskarth: if this is the same as kademlia connectivity, I'd suggest calling that out directly.

SWIPs/swip-23-kademlia.md
```diff
@@ -0,0 +1,238 @@
...
+    *MUST* attempt to send to exactly *one* peer in the bin that is the
+    closest match to the recipient address.
+
+5.  If this is not successful, the node *MUST* try to forward to each
+    remaining peer in that bin in order from closest to farthest to the
+    recipient address, until one is successful.
+
+6.  If this is not successful, the node should retain the content and
+    retry later.
+
```
oskarth: Perhaps an example section here would be useful

nagydani:
> @nagydani Regarding your bullet points on kademlia properties; are any of these properties particular to the swarm implementation? If they are all common properties, and by this I mean common in the sense that the same guarantees are provided by the Kademlia Paper by Maymounkov/Mazieres, then I think it makes more sense to merely refer to that paper and say something like:
> 
> > The Swarm implemenation of Kademlia provides the usual guarantees of logarithmic proportions regarding connectivity and routing.

The first two are common to all Kademlia variants and are, indeed, mentioned in the Maymounkov/Mazieres paper. The third one is particular to Swarm's flavor of forwarding Kademlia, though I do not claim that it is unique to it. Unfortunately, there is a lot less literature on forwarding Kademlias than the classical ones. In fact, it depends on the peer selection strategy and might not even necessarily be true with the current implementation. It actually needs some research.

nolash:
@nagydani ok does that mean that we do as I suggest for the two first points, and attach your point of "betweenness" to it? Could we word the point slightly more layman friendly, you think?

PR #27, opened 2019-09-14
SPEC: Chunks and content hashing

nolash:
**This is work in progress.**

The full evolving spec document can be viewed here:

* (latex source) https://github.com/ethersphere/node-implementer-spec
* (pdf) https://swarm-gateways.net/bzz:/dcad67ce1b19aa80d8bb2c1e5c2244c2e67f3f0f894cf863ed5522d0943e4c14/

### DIFF
--- a/SWIPs/swip-chunk-and-hasher.md
+++ b/SWIPs/swip-chunk-and-hasher.md
@@ -155,8 +155,8 @@ actual valid address that can be retrieved from the store.
 
 The symmetric key provided by the caller when adding the content is used
 as the seed to derive all decryption keys for the `Intermediate Chunks`
-[0.4.1](#sec:intermediate-chunks). It is also used as the encryption key
-for the `Swarm Hash` [sec:swarm-hash](https://github.com/nolash/SWIPs/blob/hashing/SWIPs/swip-chunk-and-hasher.md#swarm-hash).
+[0.4.1](#intermediate-chunks). It is also used as the encryption key for
+the `Swarm Hash` [0.4](#swarm-hash).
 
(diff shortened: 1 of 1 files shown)


SWIPs/swip-chunk-and-hasher.md
```diff
@@ -0,0 +1,129 @@
...
+
+N/A
+
+## Copyright
+
+Copyright and related rights waived via CC0
+
+## Specification
+
+Data stored in Swarm consists of \texttt{Chunks}.
```
oskarth: leftover latex formatting?

SWIPs/swip-chunk-and-hasher.md
```diff
@@ -0,0 +1,129 @@
...
+
+\subsection{Chunk Size}
+
+The size of a chunk is the digest size of the \texttt{Base Hash} multiplied by an arbitary exponential of 2. Currently the size is defined as 4096 bytes \footnote{$2^7 * 32$}
```
oskarth: if this is just a constant and can be defined as such with a standard definition field

this is needlessly hard to parse imo, e.g. compare with something like https://github.com/ethereum/eth2.0-specs/blob/dev/specs/simple-serialize.md#constants which is easier to scan without verbiage

oskarth: if we want to have motivations or rational for picking certain constants perhaps they can be part of separate section at bottom, similar to https://github.com/ethereum/eth2.0-specs/blob/dev/specs/networking/p2p-interface.md#design-decision-rationale

PR #28, opened 2019-09-14
Kademlia load balancing

nolash:
**in progress**

### DIFF
--- a/SWIPs/swip-kademlia-load-balancing.md
+++ b/SWIPs/swip-kademlia-load-balancing.md
@@ -1,0 +1,57 @@
+---
+SWIP: <to be assigned>
+title: Kademlia load balancing
+author: lash <dev@holbrook.no> (https://holbrook.no) Álvaro <alvaro@epiclabs.io>
+discussions-to: <URL>
+status: Draft
+type: <Standards Track (Core, Networking, Interface)  | Meta | Informational>
+category (*only required for Standard Track): <Core | Networking | Interface | ERC>
+created: <date created on, in ISO 8601 (yyyy-mm-dd) format>
+requires (*optional): <SWIP number(s)>
+replaces (*optional): <SWIP number(s)>
+---
+
+<!--You can leave these HTML comments in your merged SWIP and delete the visible duplicate text guides, they will not appear and may be helpful to refer to if you edit it again. This is the suggested template for new SWIPs. Note that a SWIP number will be assigned by an editor. When opening a pull request to submit your SWIP, please use an abbreviated title in the filename, `SWIP-draft_title_abbrev.md`. The title should be 44 characters or less.-->
+This is the suggested template for new SWIPs.
+
+Note that a SWIP number will be assigned by an editor. When opening a pull request to submit your SWIP, please use an abbreviated title in the filename, `SWIP-draft_title_abbrev.md`.
+
+The title should be 44 characters or less.
+
+## Simple Summary
+
+Initial improvement for evenly distributing requests among peers.
+
+## Motivation
+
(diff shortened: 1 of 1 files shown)


SWIPs/swip-kademlia-load-balancing.md
```diff
@@ -0,0 +1,57 @@
...
+
+## Backwards Compatibility
+
+The implementation should have no compabitility issues
+
+## Test Cases
+
+* Given `m` requests to `n`  peers in a bin, after `m*n` requests the peers will have made `m` requests each.
+* Given `m` requests to `n` peers within the neighborhood, after `m*n` requests the peers will have made `m` requests each.
+* Given `m` requests for no capability and `ma` requests for capability `a` to peers `n` with all capabilities and peers `na` with capability `a`, after `n*n + ma*na` requests the `ma` peers will have made `ma+m` requests each.
```
kortatu: n*n -> m*n
the `ma` -> the `na`

I think it should be.
```suggestion
* Given `m` requests for no capability and `ma` requests for capability `a` to peers `n` with all capabilities and peers `na` with capability `a`, after `m*n + ma*na` requests the `na` peers will have made `ma+m` requests each.
```

SWIPs/swip-kademlia-load-balancing.md
```diff
@@ -0,0 +1,57 @@
...
+
+## Motivation
+
+Currently no logic exists to guarantee that routing is distributed among peers in a bin. If a bin is unchanged between kademlia queries for forwarding peers in that bin, currently the most likely result is that the same peer will always be returned. This 
+
+Hence a load balancing mechanism should be introduced. Load balancing is a topic that can be made as complex as desired. However, a simple round-robin scheme distributing requests evenly among the peers is already a vast improvement over the current (lack of) scheme.
+
+## Specification
+
+The component should be a layer above the kademlia itself, and most likely sits in the same layer as - or is even perhaps the same component as - the planned convergence of the request and message forwarding engine (todo link to related issue).
```
kortatu: Forwarding issue: 
https://github.com/ethersphere/swarm/issues/1656

kortatu:
What is the definition of a request? Should be accounted each time a peer is returned to the caller or it depends on the use that caller is doing with the Peer.
For example, in Pss, not all peers retrieved with `EachConn` are used:
```go
p.Kademlia.EachConn(to, addressLength*8, func(sp *network.Peer, po int) bool {
		if po < broadcastThreshold && sent > 0 {
			return false // stop iterating
		}
...
```
In this case, sp won't be used.
I can see here two possibilities:

1. Each time a peer is used to execute the f function passed to EachConn is accounted as one request.

1. The caller to the load balancing service must signal in some way that a Peer has been used (complicating the API of that service) to be accounted as one request.

nolash:
Number 2.

A bool argument should suffice, I guess? We can rename the function, then have a new function with the existing name wrap that function, with that argument set so accounting is not done.

kortatu:
> Number 2.
> 
> A bool argument should suffice, I guess? We can rename the function, then have a new function with the existing name wrap that function, with that argument set so accounting is not done.

I was implementing exactly that, but is a returning value of that f function (now returns two values, one to continue iterating one to signaling that the peer has been used)

SWIPs/swip-kademlia-load-balancing.md
```diff
@@ -0,0 +1,57 @@
...
+
+The title should be 44 characters or less.
+
+## Simple Summary
+
+Initial improvement for evenly distributing requests among peers.
+
+## Motivation
+
+Currently no logic exists to guarantee that routing is distributed among peers in a bin. If a bin is unchanged between kademlia queries for forwarding peers in that bin, currently the most likely result is that the same peer will always be returned. This
```
acud: "returned. This
Hence ... "
can you please reword this?

PR #33, opened 2019-10-24
bzz-retrieve protocol spec

acud:
Long overdue but here it is. Ready for a first round of reviews


acud:
@nagydani ping

acud:
does this mean that this is bound to get merged at some point?

PR #44, opened 2024-02-12
Swip increase reserve

crtahlin:
SWIP about increasing reserve size of nodes, allowing more and relatively cheaper storage capacity.


significance:
obsoleted by SWIP-21

PR #45, opened 2024-03-02
Fix BZZ economics

jpritikin:


0xCandy:
Thanks for your hardwork and contributions! It is a great job!

0xCandy:
"Send 10% of the BZZ stamp payment to the Swarm Foundation owned Uniswap v2 DAI-BZZ pool before sending the remainder to the redistribution lottery." - BZZ stamp is bought on Gnosis, but sent to Ethereum?

jpritikin:
> "Send 10% of the BZZ stamp payment to the Swarm Foundation owned Uniswap v2 DAI-BZZ pool before sending the remainder to the redistribution lottery." - BZZ stamp is bought on Gnosis, but sent to Ethereum?

This is use refinement from a developer. I am agnostic about whether it happens on Gnosis or the Ethereum main chain.

tonytony32:
Insightful writeup. I have a couple of questions : 
- Regarding token supply: If the BC is effectively shutdown, will BZZv2 have a fixed supply?
- The governance of the UniV2 contract for SF is a bit tricky, in my view. For it to be effectively run by the community, some sort of governance contract must be in place. Is there any example you see that it'd fit properly?

jpritikin:
>     * Regarding token supply: If the BC is effectively shutdown, will BZZv2 have a fixed supply?

With the bonding curve shutdown, no additional BZZv1 can be minted. Therefore, the supply of BZZv2 is limited by the number of BZZv1 tokens minted. Moreover, BZZv2 that is sent to the Uniswap LP can be considered burnt. Therefore, the BZZv2 supply becomes deflationary.

>     * The governance of the UniV2 contract for SF is a bit tricky, in my view.

Why? What kind of governance would be needed? An advantage of UniV2 is that it don't require any active management. The only reason that I think it should be owned by the Swarm Foundation is for emergencies or if the foundation decides to change something in the distant future.

> For it to be effectively run by the community, some sort of governance contract must be in place. Is there any example you see that it'd fit properly?

Maker has a governance system, but it is very cumbersome. If at all possible, I'd prefer if we could limit the scope of governance to choices that do not have a big impact and allow the Swarm Foundation to make the decisions.

PR #47, opened 2024-05-09
Added SWIPs 19 (neighbourhood hopping) and 20 (staking change)

dysordys:


dysordys:
---
SWIP: 19
title: Neighbourhood hopping
author: György Barabás <gyorgy@ethswarm.org> (@dysordys), Viktor Trón <viktor@ethswarm.org> (@zelig)
discussions-to: https://discord.gg/Q6BvSkCv
status: Draft
type: Standards Track
category: Core
created: 2024-05-08
---


## Abstract

This SWIP describes how to support **neighbourhood hopping** (moving a node to other neighbourhoods) without losing one's stake, i.e., the same node may change its overlay address while preserving the stake it already had.


## Objectives

Currently, the staking smart contract associates a stake with an overlay address. Therefore, neighbourhood hopping requires the node operator to re-stake their node. This is considered unfair and against the spirit of the system, because the ability to change neighbourhoods is a prerequisite for the balanced distribution of nodes in the network.


## Context

Neighbourhood hopping is currently supported by the bee client. The overlay address (which specifies which neighbourhood the node falls into) is derived from the node key by hashing it together with an arbitrary overlay nonce chosen by the user (see figure below). 


Screenshot 2024-05-09 at 2.56.26 PM

Bee already supports putting one's node in an arbitrary neighbourhood by mining an overlay nonce that will result in the corresponding neighbourhood.


## Specification

To support neighbourhood hopping with transferrable stakes, the staking contract should register the user's Ethereum address, not their overlay address.


## Implementation notes

The implementation requires two indices: the Ethereum address and also the overlay address (as before). The staking API already requires the sender to send the overlay address of the nodes to calculate the overlay. Therefore, there is no change in the interface functions. However, their semantics will be changed: now, in the case of a new overlay nonce, the old node overlay address should be deleted.

The stake endpoint:
1. Look up stake in the index by Ethereum address.
2. If not found, insert it in the same index with overlay calculated from the address and the nonce sent as part of transaction data. Then go to 4.
3. If found, remember old overlay, and update the entry of the Ethereum address with it into the same index.
4. Remove the old entry in the other index (keyed by overlay), and insert new overlay value updated with additional funds.


## Backward compatibility

Since we are changing how information is passed (both the overlay address and the Ethereum address), this change is not backwards-compatible.


## Test cases

Case study, with individual items needing to be tested:
1. Can one stake a node?
2. Is it then possible to change neighbourhoods?
3. If so, then did that happen with the stake transferred without a change?


## Copyright

Copyright and related rights waived via [CC0](https://creativecommons.org/publicdomain/zero/1.0/).

dysordys:
---
SWIP: 20
title: Improved staking - allowing excess stake withdrawal
author: György Barabás <gyorgy@ethswarm.org> (@dysordys), Viktor Trón <viktor@ethswarm.org> (@zelig)
discussions-to: https://discord.gg/Q6BvSkCv
status: Draft
type: Standards Track
category: Core
created: 2024-05-08
---


## Abstract

Currently, one's initial capital, invested as stake, is permanently locked. This means that returns on this stake is purely based on the ROI of the redistribution game. The purpose of this SWIP is to change that: in extenuating circumstances, users will be allowed to withdraw their stakes.


## Objectives

Should the effective price of a user's stake increase, they should be able to withdraw it, as it is no longer representative of the capital required to play and safeguard the system.


## Context

Currently, the smart contract does not allow changing one's stake. This is what will be changed with this feature.


## Specification

This part describes how stakes should be computed, withdrawn, and returned if needed. As a first step, we need to obtain the the *committed stake* of the user. This is set by user action, based on the following procedure:

1.  The stake endpoint of the contract accepts an extra argument, `committedStake`, which sets the committed stake.
2.  The contract asks the price oracle the unit price of storage -- let us denote it with the symbol `unitPrice`.
3.  The contract then adds whatever amount is sent with the transaction to the `txorigins` stake entry. We can call this sum the *potential stake balance*, and denote it with the symbol `potentialStakeBalance`.
4.  The contract then calls the `withdraw` routine.

The `withdraw` routine, in turn, comprises the following steps:
1.  Calculate the *effective stake* as:
    ```
    effectiveStake = min(committedStake * unitPrice, potentialStakeBalance)
    ```
2.  Calculate the *surplus stake* as `surplusStake = potentialStakeBalance - effectiveStake`
3.  If `surplusStake > 0` then send `surplusStake` back to `txorigin`


## Implementation notes

- We are changing the contracts, specifically the staking contract, to attribute stake to the address and not the overlay. This will allow stake to be moved when changing neighbourhoods, and allow nodes to migrate and distribute themselves across the network.
- The contract change will involve a migration process of pausing the staking contract, which allows users to withdraw their stake and then reinvest in the new staking contract with new parameters to move stake freely when changing neighbourhoods. This will need a user-story testing session of the process and steps.
-  BEE team aspect: for staking changes, the bee node should be made to work with new ABI for contracts. This means some different function names and different variable types/names.
- BEE team aspect for withdraw: for withdrawal, some way should be added for users to withdraw funds for this first one (and for future ones if there isn't one).


## Backward compatibility

This is not going to be backward compatible. Question: would currently staked nodes be able to withdraw after the upgrade, or would they need to re-stake, with the previous stake lost?


## Test cases

1. Can one set up a node?
2. Can one stake it?
3. Can one top it up with progressively more BZZ?
4. Once that is done, can one withdraw funds from that stake?

A question for an imagined worst-case scenario: if BZZ increases incrementally, which incentivizes nodes to withdraw funds, and this happens simultaneously for many nodes, could that put the network as a whole in danger?


## Copyright

Copyright and related rights waived via [CC0](https://creativecommons.org/publicdomain/zero/1.0/).

zelig (approved):

PR #48, opened 2024-05-13
Add SWIP-19: Neighbourhood hopping

dysordys:
This SWIP describes how to support neighbourhood hopping; i.e., moving a node to another neighbourhood without the need to pay the stake each time.

### DIFF
--- a/SWIPs/swip-19.md
+++ b/SWIPs/swip-19.md
@@ -1,0 +1,139 @@
+---
+SWIP: 19
+title: Neighbourhood hopping
+author: György Barabás <gyorgy@ethswarm.org> (@dysordys), Viktor Trón <viktor@ethswarm.org> (@zelig), Mark Bliss <marko@ethswarm.org>
+discussions-to: https://discord.gg/Q6BvSkCv
+status: Draft
+type: Standards Track
+category: Core
+created: 2024-05-08
+---
+
+## Abstract
+
+This SWIP describes the process to support **neighbourhood hopping**, i.e., moving a node to another neighbourhood without the need to pay the stake each time.
+
+## Objectives
+
+Enable a node to change its overlay address while preserving the stake it already possesses.
+
+## Context
+
+Neighbourhood hopping is currently supported by the Bee client. The overlay address, which specifies which neighbourhood a node falls into, is derived by hashing the node key together with an arbitrary overlay nonce chosen by the user (refer to the figure below).
+
+![](assets/swip-19/overlay-definition.png)
+
+Bee already supports placing one's node in an arbitrary neighbourhood by mining an overlay nonce that will result in the corresponding neighbourhood.
+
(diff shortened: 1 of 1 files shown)


SWIPs/swip-19.md
```diff
@@ -1,10 +1,132 @@
...
+
+This SWIP describes the process to support **neighbourhood hopping**, i.e., moving a node to another neighbourhood without the need to pay the stake each time.
+
+## Objectives
+
+Enable a node to change its overlay address while preserving the stake it already possesses.
+
+## Context
+
+Neighbourhood hopping is currently supported by the Bee client. The overlay address, which specifies which neighbourhood a node falls into, is derived by hashing the node key together with an arbitrary overlay nonce chosen by the user (refer to the figure below).
```
janos, line 22: `Neighbourhood hopping is currently supported by the Bee client.` is a questionable statement. As mentioned in this parameter, it is possible to mine an overlay address that falls into a specific neighbourhood, but neighbourhood hopping has to be defined in the terms of the feature requirements. As the overlay address changes, syncing indexes and intervals in the database need to be adjusted, which will create a noticeable impact on the usability and performance of the node. I am not aware of an automated way to reindex syncing, either at the runtime or before the node is started. This has to be implemented on regards when neighbourhood hopping can happen, at the runtime or on start. Alternative is to remove all node storage, but is not ideal.

janos (commented):
This is a good proposal, I am just raising awareness of the impact from the Bee node implementation aspect.

SWIPs/swip-19.md
zelig: Thank you janos for raising these points.

Indeed how the node is supposed to do neighbourhood hopping should be discussed and written.
But I am not sure the details should part of this SWIP, rather they belong to documention.

zelig: very good point about syncing and frequent hoppers. 
unfortunately the frequent hopping attack can be played the same without stake, so this swip made no difference in that.
Mitigating against this attack must be part of the planned pull-sync upgrade SWIP.  That proposal aspires to optimise the sync protocol by eliminating predictable redundancies that are the cause of excessive bandwidth.

janos: @zelig thanks for addressing these concerns. I agree that the documentation is the right place for details about the neighbourhood hopping procedure on the UX and internals. It is the case that current state of the implementation does not have a clear interface to perform the hopping from the UX perspective, by my opinion.

I support the effort to harden the syncing protocol in regards of the frequent hopping attack. That is indeed possible now, but I believe it would be easier to accomplish (or just be more obvious) when neighbourhood hopping is implemented and advertised.

nugaon (commented):
LGTM, it was strange for me first to have 2 events for the same change but for gas saving purposes it is fine.
The Bee code could change towards better UX at any time, than that may be rolled out first wit nuking. Would that be enough if initial implementation is referenced in the doc and other subsequent PRs just referenced on GitHub?

zelig (approved):

PR #49, opened 2024-05-13
Add SWIP-20: Improved staking

dysordys:
Currently, one's initial capital, invested as stake, is permanently locked. This means that returns on this stake is purely based on the ROI of the redistribution game. The purpose of this SWIP is to change that: in extenuating circumstances, users will be allowed to withdraw their stakes.


SWIPs/swip-20.md
```diff
@@ -8,3 +8,118 @@ type: Standards Track
...
+## Context
+
+Currently, the smart contract can only increase a node's stake by sending extra collateral. Importantly, it does not allow node operators to  decrease their stake or any other way fully or partially withdraw their stake. This is what will be changed with this feature.
+
+
+## Specification
+
+This part describes how stakes should be created, modified, computed and withdrawn. As a first step, we need to obtain the *committed stake* of the user. This is set by user action, based on the following procedure:
+
+### The stake endpoint
```
nugaon, line 32: this endpoint naming is confusing for me. would it be the currently existing `depositStake` function? or the Staking contract itself?

SWIPs/swip-20.md
```diff
@@ -8,3 +8,118 @@ type: Standards Track
...
+1.  The contract accepts an extra argument, `committedStake`, which sets the committed stake.
```
nugaon, line 33: where does the contract accept extra argument and what is `commitedStake`?

SWIPs/swip-20.md
```diff
@@ -8,3 +8,118 @@ type: Standards Track
...
+2.  The contract asks the price oracle for the unit price of storage -- denoted here as `unitPrice`.
```
nugaon, line 34: what does it do with the `unitPrice` in this process?

nugaon (commented):
the introduced terms could be defined in a short glossary or detailed more in place.

SWIPs/swip-20.md
zelig: ú-yes I mean the depositStake function. But talking from a bee perspective this is used as the backend of the `/stake` Bee Api.

SWIPs/swip-20.md
```diff
@@ -8,3 +8,114 @@ type: Standards Track
...
+
+Currently, the smart contract can only increase a node's stake by sending additional collateral. Importantly, it does not allow node operators to  decrease their stake or any other way to fully or partially withdraw their stake. This limitation is what will be changed with this feature.
+
+
+## Specification
+
+This section describes how stakes should be created, modified, computed, and withdrawn. As a first step, we need to determine the *committed stake* of the user. This is set by user action, based on the following procedure:
+
+### The stake endpoint
+1.  The contract accepts an extra argument, `committedStake`, which sets the committed stake.
```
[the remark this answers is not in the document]
zelig: > Is the committed stake the same as the current stakeAmount we have, if so why do we say new argument?

No it is not.

SWIPs/swip-20.md
```diff
@@ -8,3 +8,114 @@ type: Standards Track
...
+2.  The contract asks the price oracle for the unit price of storage -- denoted here as `unitPrice`.
+3.  The contract then adds the amount sent with the transaction to the stake entry for the address of `txorigin`. This sum can be referred to as the *potential stake balance*, denoted as `potentialStakeBalance`.
```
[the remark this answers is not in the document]
zelig: added a concepts section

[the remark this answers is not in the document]
n00b21337: much clearer now, thank you

nugaon (commented):
It is not completely clear for me why `commitedStake` is a parameter instead of just being calculated in the `depositStake` function. Is there any reason to set lower `commitedStake` that does not fully cover the sent BZZ amount?

SWIPs/swip-20.md
```diff
@@ -8,3 +8,139 @@ type: Standards Track
...
+
+
+## Abstract
+
+Currently, a storer node's stake to buy their way into the redistribution game, is permanently locked which means that returns on this stake are solely based on the ROI of the redistribution game. The purpose of this SWIP is to change this and allow users to withdraw part of their stakes.
+
+
+## Objectives
+
+Should the effective price of a user's stake increase, they should be able to withdraw the appreciation, i.e., the portion of the value that is above the amount required to play and safeguard the system.
```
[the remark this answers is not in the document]
n00b21337: I will model some examples here with amounts.

Step 1
100 BZZ added

potential is
100000000000000000  at 25000 price
committed is 40000000000000

Step 2

when the price of oracle goes to 26000
committed will be 1040000000000000000

so there is no surplus as surplusStake = potentialStakeBalance - effectiveStake 
means it is basically 0 as effective stake is min(pot,com) which is the current potential
so user cant take out anything.

Step 3
when the price of oracle goes to 24000
then committed will be 960000000000000000

so there will be a surplus of 40000000000000000 which is 4 BZZ and he can take it out

Step 4
If the price oracle goes up or is the same, there is no option to take out funds lets use the same price
committed is 960000000000000000000
potential is the same as we took out 4 BZZ
so until priceOracle goes up, node cant take out anything, it always needs to go down for the next withdrawal

Step 5
price oracle goes to 23000
committed is 40000000000000  x 23000   = 920000000000000000
potential is 960000000000000000

he can again take out 4BZZ

SWIPs/swip-20.md
```diff
@@ -17,7 +17,7 @@ Currently, a storer node's stake to buy their way into the redistribution game,
 
 ## Objectives
 
-Should the effective price of a user's stake increase, they should be able to withdraw the appreciation, i.e., the portion of the value that is above the amount required to play and safeguard the system.
+Should the effective price of a user's stake dencrease, they should be able to withdraw the appreciation, i.e., the portion of the value that is above the amount required to play and safeguard the system.
```
zelig, line 20: no. if the value increases:

Should the valued of a user's stake increase, they should be able to withdraw the appreciation

SWIPs/swip-20.md
```diff
@@ -93,7 +94,7 @@ We are modifying the contracts, specifically the staking contract, to attribute
 
 ### Swarm API changes
 - *API Stake endpoint:*
-   For the staking node, the value for committed stake should be solicited as user input in the form of an added request parameter.
+   For the staking node, the parametars that are set are same as in swip-19, which is sending value and nonce
```
zelig, line 97: parametErs

zelig (changes requested):

SWIPs/swip-20.md
n00b21337: I already asked about this above, can we check the math as it clearly says

**surplusStake = potentialStakeBalance - effectiveStake**

which means effective stake needs to get lower to be able to take funds out. Both Nugaon and me concluded that  effective stake needs to be lower. If we are not talking here about effective Stake but just price of committed stake in BZZ again it needs to get lower as we are substracting from min(pot,com).
I do wonder if something is wrong here with math or wording . @nugaon  please also jump in here

nugaon: agree as we discussed.
since then, I see the misleading `effective price` has been changed to `value` which is correct.

PR #53, opened 2024-06-06
New SWIP process - complete remake

zelig:
just contains new proposal as an update to README.

If accepted, should follow with complience changes

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,34 +1,66 @@
-# SWIPs
-Swarm Improvement Proposals (SWIPs) describe standards for the Swarm platform, including core protocol 
-specifications, client APIs, and contract, interface and utility standards.
+ <div align="center">
+<h1 align="center">SWIPs</h1>
+<p align="center">
+    <a href="https://discord.com/channels/799027393297514537/1239813439136993280">
+        <img src="https://img.shields.io/badge/Community%20Channel%20for%20SWIPs-white?logo=discord">
+    </a>
+</p>
+<h3 align="center">Swarm Improvement Proposal repository.</h3>
+</div>
+
+<div align="center">
+<br />
+</div>
+
+<details open="open">
+<summary>Table of Contents</summary>
+
+- [About](#about)
+- [Contributing](#contributing)
+  - [Creating a new SWIP](#creating-a-new-swip)s
+  - [Updating a SWIP](#updating-a-swip)
+  - [Format and Structure](#format-and-structure)
+- [Curating SWIPs](#curating-swips)
+  - [SWIP Status Terms](#swip-status-terms)
+
+</details>
+
+# About
+
+Swarm Improvement Proposals (SWIPs) describe standards for the Swarm platform, including core protocol specifications, client APIs, and standards for contracts, interfaces, and utilities.
 
 A browsable version of all current and draft SWIPs can be found on [GitHub](SWIPs).
-<!-- TODO: This must be Swarm site eventually -->
 
 ## Contributing
 
-This README contains a high level description of contributing a SWIP, for convenience.
+This README provides a high-level description of the process for contributing a SWIP.
 
- 1. Review [SWIP-0](SWIPs/swip-0.md).
- 2. Fork the repository by clicking "Fork" in the top right.
- 3. Add your SWIP to your fork of the repository. There is a [template SWIP here](SWIPs/swip-X.md).
- 4. Submit a Pull Request to Swarm's [SWIPs repository](https://github.com/ethersphere/SWIPs).
+1. Review [SWIP-0](SWIPs/swip-0.md).
+2. Fork the repository by clicking "Fork" in the top right.
+3. Add your SWIP to your fork of the repository. A template SWIP can be found [here](SWIPs/swip-X.md).
+4. Submit a Pull Request (PR) to the Swarm [SWIPs repository](https://github.com/ethersphere/SWIPs/pulls).
 
-A SWIP proposal or suggested edits (or updates) may come from anyone (inside or outside the organisation) and are received in the form of a pull request to the [SWIP repository](https://github.com/ethersphere/SWIPs). Throughout the lifetime of a SWIP from proposal to final accepted, we use the PR-review infrastructure of github to comment and improve the SWIP.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -14,30 +14,76 @@ This README contains a high level description of contributing a SWIP, for conven
...
 
-* **Draft** - a SWIP that is undergoing rapid iteration and changes.
-* **Last Call** - a SWIP that is done with its initial iteration and ready for review by a wide audience.
-* **Accepted** - a core SWIP that has been in Last Call for at least 2 weeks and any technical changes that were requested have been addressed by the author. The process for Core Devs to decide whether to encode a SWIP into their clients as part of a mandatory update is not part of the SWIP process. If such a decision is made, the SWIP will move to final.
-* **Final (non-Core)** - a SWIP that has been in Last Call for at least 2 weeks and any technical changes that were requested have been addressed by the author.
-* **Final (Core)** - a SWIP that the Core Devs have decided to implement and release in a future mandatory update or has already been released as such. 
-* **Deferred** - a SWIP that is not being considered for immediate adoption. May be reconsidered in the future.
+```yaml
+ ---
+title: fairx
```
crtahlin, line 32: This example title is concise, but not descriptive. I suggest taking one of the existing as example:
Neighbourhood hopping

README.md
```diff
@@ -14,30 +14,76 @@ This README contains a high level description of contributing a SWIP, for conven
...
+- Only SWIP editors can merge a PR to master.
 
-When you believe your SWIP is mature and ready to progress past the draft phase, you should open a PR changing the state of your 
-SWIP to 'Final'. An editor will review your draft and ask if anyone objects to its being finalised. If the editor decides there 
-is no rough consensus - for instance, because contributors point out significant issues with the SWIP - they may close the PR and 
-request that you fix the issues in the draft before trying again.
+### Creating a new SWIP
 
-For a detailed description of the SWIP process see [SWIP-0](SWIPs/swip-0.md).
+A new SWIP can be proposed by opening a PR in the [SWIP repo](https://github.com/ethersphere/SWIPs) titled `SWIP-<XXXX>: <title>` where `XXXX` is the SWIP number left padded with zeros up to four digits (i.e., fprinted with the format string `%04d`).
```
crtahlin, line 25: To avoid clashes on SWIP numbers, let us define that a SWIP must have the same number as the PR. 

Also, I would be against left padding with zeros to make it for digits, as I see no need for that. The number could quickly grow to more than four digits. And also, the practice in Ethereum is there is no padding (example https://eips.ethereum.org/all).

README.md
```diff
@@ -14,30 +14,76 @@ This README contains a high level description of contributing a SWIP, for conven
...
+The SWIP editors can merge this initial request to signal the SWIP worthy of recording and ready for discussion.
+
+From this moment on, the SWIP number is taken as allocated and the SWIP will permanently be part of the repo (yes even if it is deferred or rejected).
+This way looking at the repo should give the number to be used when  a new SWIP proposal is submitted. 
+
+Submitters of SWIPs are recommended to use a mnemonic feature branch name. 
+
+### Updating a SWIP
+
+A suggested edit to a SWIP can be proposed by opening a PR in the [SWIP repo](https://github.com/ethersphere/SWIPs) with a PR title in the following format: `Update SWIP-XXXX: Title of the suggested change`.
```
crtahlin, line 53: This is for an edit of a SWIP that is already final / accepted? As a draft SWIP should be just edited directly?

crtahlin: Actually, reading forward, it is for a SWIP that is merged but can be in any state?

README.md
```diff
@@ -14,30 +14,76 @@ This README contains a high level description of contributing a SWIP, for conven
...
+
+### Updating a SWIP
+
+A suggested edit to a SWIP can be proposed by opening a PR in the [SWIP repo](https://github.com/ethersphere/SWIPs) with a PR title in the following format: `Update SWIP-XXXX: Title of the suggested change`. 
+
+When a PR is already open for the SWIP improvments must be submitted as PRs against this feature branch. Organising reviews to these and merging them is delegated to the submitter (names in the `authors` header).
+
+Once a phase of a SWIPs is complete, authors will indicate this by setting the `ready to merge` label on the open PR. editors can merge the PR to master. see also section about curating SWIPs below.
+
+### Format and structure
```
crtahlin, line 59: Should the template SWIP (https://github.com/ethersphere/SWIPs/blob/master/SWIPs/swip-X.md) still be used? I guess not, since there are sections and headers defined in this README.

README.md
zelig: OK then. Then it is no longer relevant to merge the claim PR early on, it simplifies the process.

The need for XXXX numbering is solely the neat temporally correct listing of SWIP md files

zelig: @crtahlin How about we don't use /assets/SWIP-XXXX/ subdirectories, rather a global assets directory to enable shared figures?

nikipapadatou: I would say that each SWIP should have their own folder of assets for easier reference and browsing within the SWIP. If in the future we see the need for having shared assets among the SWIPs, we can have a global one.

README.md
```diff
@@ -14,30 +14,76 @@ This README contains a high level description of contributing a SWIP, for conven
...
 
-When you believe your SWIP is mature and ready to progress past the draft phase, you should open a PR changing the state of your 
-SWIP to 'Final'. An editor will review your draft and ask if anyone objects to its being finalised. If the editor decides there 
-is no rough consensus - for instance, because contributors point out significant issues with the SWIP - they may close the PR and 
-request that you fix the issues in the draft before trying again.
+### Creating a new SWIP
 
-For a detailed description of the SWIP process see [SWIP-0](SWIPs/swip-0.md).
+A new SWIP can be proposed by opening a PR in the [SWIP repo](https://github.com/ethersphere/SWIPs) titled `SWIP-<XXXX>: <title>` where `XXXX` is the SWIP number left padded with zeros up to four digits (i.e., fprinted with the format string `%04d`). 
+The PR is adding at least a new placeholder file for the new proposal named `XXXX.md` to be located in the `SWIPs` subdirectory of the SWIPs repo.
```
nikipapadatou, line 26: Let's also define the name format here, like: swip-xx.md (where xx is the number of the PR, as discussed above)

tamas6:
Please review @zelig 🐝

PR #61, opened 2024-11-07
SWIP-33 time-based-incentives-weights

darkobas2:


tamas6:
Thanks for this SWIP proposal! The idea of rewarding faster nodes is compelling.

based on the bee dev discussions here is a quick summary:

- The current reward system is deterministic and tied to expired batches, making time-based adjustments complex.
- Speed primarily impacts the probability of selection, not the reward amount directly.
- Prioritizing CPU performance may unfairly penalize normal nodes with adequate resources.

Your initial tests with modern CPUs show improved performance with more cores, supporting the logic. 
This aligns with Swarm's goals. But this is now a discussion opener @ct-ctrl what do you think?

PR #65, opened 2025-01-28
SWIP-24: support communication through websockets

lat-murmeldjur:
This Swarm Improvement Proposal attempts to address enabling browser-based libp2p nodes connectivity to the network and sum up the minimal set of required changes to the protocols involved in connectivity.


lat-murmeldjur:
Update: With ports consistently reused between tcp and websocket underlays we can achieve derivation of websocket underlays from gossiped tcp underlays, that enables hive protocol to remain intact

lat-murmeldjur:
Update 2: It seems port reuse has some issues that make it less feasible as a long term solution, with requests hanging for 10 seconds periodically. Reopening this as a mid-term solution proposal

significance:
being worked on here https://github.com/ethersphere/bee/pull/5187

it would be good to align the specification here with that implementation

lat-murmeldjur:
> being worked on here [ethersphere/bee#5187](https://github.com/ethersphere/bee/pull/5187)
> 
> it would be good to align the specification here with that implementation

Looking for a way to find bootnode wss underlay on testnet

PR #67, opened 2025-03-03
add swip-26: strictly typed chunk system

mfw78:
This PR introduces a draft SWIP for a standardised framework for defining chunk types in Swarm, improving security and interoperability through consistent type identification and validation.

### DIFF
--- a/SWIPs/swip-26.md
+++ b/SWIPs/swip-26.md
@@ -110,6 +110,24 @@ The chunk processing logic shall:
 
 This approach allows for early validation of chunk integrity based on protocol-level type information, reducing parsing errors and simplifying processing logic.
 
+#### Flowchart
+
+The flowchart below illustrates the processing steps for a chunk:
+
+```mermaid
+flowchart TD
+    Start[Receive chunk via wire protocol] --> A[Protobuf decodes chunk type, version, header, and payload]
+    A --> B{Header size matches expected size for type?}
+    B -->|No| C[Fail: Invalid header size]
+    B -->|Yes| D[Extract type-specific header fields]
+    D --> E[Calculate chunk address using type-specific function]
+    E --> F{Validate chunk content}
+    F -->|Invalid| G[Fail: Invalid chunk content]
+    F -->|Valid| H[Process payload according to type-specific structure]
+    H --> I[Pass processed chunk to appropriate protocol handler]
+    I --> End[Protocol-specific processing]
+```
+
 ## Rationale
 
 The proposed standardised chunk type framework addresses several key issues in the current implementation:
--- a/SWIPs/swip-26.md
+++ b/SWIPs/swip-26.md
@@ -12,10 +12,10 @@ created: 2025-03-03
 This SWIP introduces a standardised framework for defining chunk types in Swarm, improving security and interoperability through consistent type identification and validation.
 
 ## Abstract
-This SWIP proposes a standardised framework for defining and processing chunk types in Swarm. By creating a formal type system for chunks, including content-addressed chunks (CAC) and single-owner chunks (SOC), we improve security, interoperability, and maintainability across the Swarm ecosystem. The proposal defines a structured approach to chunk identification, versioning, and validation without modifying the wire protocol. Key innovations include fixed-length type-specific headers, deterministic address calculation, and formalised validation rules.
+This SWIP proposes a standardised framework for defining and processing chunk types in Swarm. By creating a formal type system for chunks, including content-addressed chunks (CAC) and single-owner chunks (SOC), we improve security, interoperability, and maintainability across the Swarm ecosystem. The proposal defines a structured approach to chunk identification, versioning, and validation. The key innovation is the formal definition of fixed-length type-specific headers to be delivered alongside chunks and formally documenting address determination and payload validation rules.
 
 ## Motivation
-Swarm's storage layer is built around chunks as the fundamental unit of data. Currently, the system supports multiple chunk types, but lacks a standardised type system. This creates several issues:
+Swarm's storage layer is built around chunks as the fundamental unit of data. Currently, the system supports multiple chunk types, but lack standardised headers. This creates several issues:
 
 1. **Ambiguous Processing**: Without explicit type information, chunk processing depends on implicit detection methods, leading to potential security vulnerabilities.
 
--- a/SWIPs/swip-26.md
+++ b/SWIPs/swip-26.md
@@ -1,0 +1,173 @@
+---
+swip: 26
+title: Standardised Chunk Type Framework
+status: Draft
+type: Standards Track
+category: Core
+author: mfw78 (@mfw78)
+created: 2025-03-03
+---
+
+## Simple Summary
+This SWIP introduces a standardised framework for defining chunk types in Swarm, improving security and interoperability through consistent type identification and validation.
+
+## Abstract
+This SWIP proposes a standardised framework for defining and processing chunk types in Swarm. By creating a formal type system for chunks, including content-addressed chunks (CAC) and single-owner chunks (SOC), we improve security, interoperability, and maintainability across the Swarm ecosystem. The proposal defines a structured approach to chunk identification, versioning, and validation without modifying the wire protocol. Key innovations include fixed-length type-specific headers, deterministic address calculation, and formalised validation rules.
+
+## Motivation
+Swarm's storage layer is built around chunks as the fundamental unit of data. Currently, the system supports multiple chunk types, but lacks a standardised type system. This creates several issues:
+
+1. **Ambiguous Processing**: Without explicit type information, chunk processing depends on implicit detection methods, leading to potential security vulnerabilities.
+
+2. **Limited Extensibility**: Adding new chunk types requires changes to core validation logic, making it difficult to evolve the system.
+
+3. **Inconsistent Validation**: Chunk validation logic is spread across multiple components, leading to potential inconsistencies.
+
+4. **Type-Safety Gaps**: Without formal type definitions, runtime type errors can occur when processing chunks.
+
+A standardised chunk type framework would address these issues by providing a consistent, extensible system for defining, identifying, and validating different chunk types.
+
+## Specification
+
+### Core Concepts
+
+#### 1. Chunk Structure
+
+A standardised chunk shall conceptually consist of:
+
+1. **Header**: Metadata describing the chunk and its contents
+   - Common Header: Information common to all chunk types (type, version)
+   - Type-Specific Header: Additional fields specific to the chunk type
+2. **Payload**: The actual chunk data
+
+The chunk's address is not part of the chunk itself but is deterministically derived from the chunk's contents based on its type.
+
+#### 2. Common Chunk Header
+
+The common chunk header shall contain:
+
+1. **Type**: The chunk type identifier (1 byte)
+2. **Version**: The chunk format version (1 byte)
+
+| Type ID | Name | Description |
+|---------|------|-------------|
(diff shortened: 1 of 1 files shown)


SWIPs/swip-26.md
```diff
@@ -0,0 +1,191 @@
...
+- Extensibility through the standardised specification system
+- Backward compatibility with existing chunk types
+
+## Backwards Compatibility
+
+This proposal maintains backward compatibility by:
+
+1. Preserving existing chunk address calculation methods for current chunk types
+2. Supporting current chunk formats with version 0 of each type
+3. Allowing for gradual adoption of the type system
```
nugaon, line 157: for me, first, it seems like it needs a breaking change in the base protocol to handle type headers.

nugaon (commented):
The chunk address always available before [chunk type assertion](https://github.com/ethersphere/bee/blob/master/pkg/soc/validator.go#L14) which kind of implies its type; if it cannot be a SOC it is a CAC based on its address.
Though the proposal has valid points and straightforward logic to solve chunk types, I don't see the urge to incorporate this new type indicator for chunks because there is no new chunk type that needs it.

SWIPs/swip-26.md
```diff
@@ -0,0 +1,191 @@
...
+
+## Abstract
+This SWIP proposes a standardised framework for defining and processing chunk types in Swarm. By creating a formal type system for chunks, including content-addressed chunks (CAC) and single-owner chunks (SOC), we improve security, interoperability, and maintainability across the Swarm ecosystem. The proposal defines a structured approach to chunk identification, versioning, and validation without modifying the wire protocol. Key innovations include fixed-length type-specific headers, deterministic address calculation, and formalised validation rules.
+
+## Motivation
+Swarm's storage layer is built around chunks as the fundamental unit of data. Currently, the system supports multiple chunk types, but lacks a standardised type system. This creates several issues:
+
+1. **Ambiguous Processing**: Without explicit type information, chunk processing depends on implicit detection methods, leading to potential security vulnerabilities.
+
+2. **Limited Extensibility**: Adding new chunk types requires changes to core validation logic, making it difficult to evolve the system.
```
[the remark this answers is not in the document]
mfw78: Replica chunks come to mind as a natural extension, instead of slicing this into the SOC chunks. By not communicating the type of a chunk on the wire-level, this makes execution non-deterministic which leaves open potential issues surrounding resource starvation (e.g. a concerted actor can send chunks that are _always_ SOC chunks, and, on the assumption that most chunks are CAC chunks, lead to a double-spike in resource consumption as all chunks must first go through the CAC check first.

mfw78:
> The chunk address always available before [chunk type assertion](https://github.com/ethersphere/bee/blob/master/pkg/soc/validator.go?rgh-link-date=2025-03-05T11%3A09%3A03Z#L14) which kind of implies its type; if it cannot be a SOC it is a CAC based on its address. Though the proposal has valid points and straightforward logic to solve chunk types, I don't see the urge to incorporate this new type indicator for chunks because there is no new chunk type that needs it.

While there is no new chunk type, this is work that would have to be achieved  prior to this. Examples that come to mind would be system chunks (whereby postage snapshots can be distributed with some other proof mechanism attached instead of stamps).

SWIPs/swip-26.md
```diff
@@ -0,0 +1,191 @@
...
+category: Core
+author: mfw78 (@mfw78)
+created: 2025-03-03
+---
+
+## Simple Summary
+This SWIP introduces a standardised framework for defining chunk types in Swarm, improving security and interoperability through consistent type identification and validation.
+
+## Abstract
+This SWIP proposes a standardised framework for defining and processing chunk types in Swarm. By creating a formal type system for chunks, including content-addressed chunks (CAC) and single-owner chunks (SOC), we improve security, interoperability, and maintainability across the Swarm ecosystem. The proposal defines a structured approach to chunk identification, versioning, and validation without modifying the wire protocol. Key innovations include fixed-length type-specific headers, deterministic address calculation, and formalised validation rules.
```
significance:
```suggestion
This SWIP proposes a standardised framework for defining and processing chunk types in Swarm. By creating a formal type system for chunks, including content-addressed chunks (CAC) and single-owner chunks (SOC), we improve security, interoperability, and maintainability across the Swarm ecosystem. The proposal defines a structured approach to chunk identification, versioning, and validation. The key innovation is the formal definition of fixed-length type-specific headers to be delivered alongside chunks and formally documenting address determination and payload validation rules.
```

SWIPs/swip-26.md
```diff
@@ -0,0 +1,191 @@
...
+
+## Motivation
+Swarm's storage layer is built around chunks as the fundamental unit of data. Currently, the system supports multiple chunk types, but lacks a standardised type system. This creates several issues:
```
significance:
```suggestion
Swarm's storage layer is built around chunks as the fundamental unit of data. Currently, the system supports multiple chunk types, but lack standardised headers. This creates several issues:
```

SWIPs/swip-26.md
significance, line 22: not sure i agree with this (core validation logic) but yes standardising the chunk type types will make it easier moving forward and we should use SWIPs to define these

significance:
suggested some changes, otherwise happy to merge. more SWIPs should follow which define the two current chunk types and use these as a stage to establish prescriptive examples of a formal chunktype type structure with which future chunktypes shall be defined

significance:
@mfw78 have read through and i think we are gtg in spirit here

however, it would be good to a little more specific in the protocol specific implementation. my memory is we agreed to send the chunk type data specified in the protobuf messages. would be very much grateful if you are up for making a suggestion on how you would like to do so. if you are able, i will canvas opinion from the Bee team and once these details are pinned down let's merge and proceed to implementation

mfw78:
Added wire protocol representation section based on recent discussion.

Key additions:
- Defined `Chunk` protobuf message with `type`, `version`, and `payload` fields
- Specified that all protocol buffer definitions referencing chunk data MUST use the typed `Chunk` message instead of raw `bytes`
- Included `Delivery` message example showing integration with pushsync/pullsync
- Added migration path for backward compatibility with legacy messages

SWIPs/swip-26.md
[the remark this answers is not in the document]
significance: i think we concur there is enough cause to implement it regardless of current dearth of solid future plans for more chunks. so marking this as resolved

significance: the specific validation methodology is a decision for those implementing client, which i am glad will be enabled with the additional information given by the headers, even for the two currently agreed chunk types

SWIPs/swip-26.md
```diff
@@ -0,0 +1,191 @@
...
+    I --> End[Protocol-specific processing]
+```
+
+## Rationale
+
+The proposed standardised chunk type framework addresses several key issues in the current implementation:
+
+1. **Type Ambiguity**: By explicitly encoding chunk types in the header, we eliminate ambiguity in chunk processing, enhancing security and reliability.
+
+2. **Extensibility**: The formal specifications allow for future chunk types to be added in a standardised way without modifying core validation logic.
```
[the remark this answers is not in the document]
significance: agree, let's ensure we discuss this if/when new chunk types are proposed

SWIPs/swip-26.md
significance: i think it is fine for a client to opine methodology around when to determine chunk types  in the first instance

as far as the protocol is concerned, i believe protobuf fields are optional, but sending headers should become mandatory once the swarm has had time to adjust. in this way the change could be made with less disruption. let's pin down our approach. thank you for raising this @nugaon

significance:
@mfw78 🙇

SWIPs/swip-26.md
```diff
@@ -0,0 +1,242 @@
...
+
+### Wire Protocol Representation
+
+To enable typed chunks at the wire level, the following Protocol Buffer definitions shall be used:
+
+#### Chunk Message
+
+```protobuf
+message Chunk {
+  uint32 type = 1;    // Chunk type identifier (see type table)
```
acud: a note about this - not sure if having a uint32 is: a. too permissive, b. too big.
since i don't see the list of chunk types and their versions to be too much in flux, consider using protobuf enums here (one for type, one for version), and we will bear the brunt of maintaining that in long run. alternatively, having `bytes` here with `len = 1` would narrow down what is needed here. honestly i think enums would work best here. also, consider having `type` and `version` to be represented in one enum - this would probably allow to simplify downstream business logic. compare:
```go
if type == 1 {
 if version == 0 {...}
 if version == 1 {...}
}
```
with:
```go
switch type {
 case swarm.CacV1:
  ...
 case swarm.CacV2:
  ...
}
```

mfw78: I'd certainly be a fan of using enums, and would prefer this, but left this method out as there seems to have been an intentional non-committal towards firm typing. I am a little concerned that the definition of the _type_ would leak _type implementation_ details into the other. Ideally as well, could also compress down to u8 instead of u32, but the idea here was to keep the contents of the frame somewhat within a cache line for a 64 bit machine (to be fair though, I don't have firm benchmarks to support necessitating this). 

Can you propose a concrete alternative / suggestion to the protobuf(s)?

significance:
after some excruciating research i have determined that u8 in protobuf is not a thing and i believe that u32 actually ends up somehow being more efficient due to some protobuf based cleverness. it seems we can pack 2 u16's into 1 u32 at the expense of api simplicity. i do think it is worth keeping the version and type separate as i can foresee situations where the type is of importance independent of the version


[summary of u32 v. bytes research for the brave](https://gist.github.com/significance/e5f5819d7d98044d770babaf52a9a00d)

acud:
@significance indeed since `u32` uses varint encoding, even "paying" a few bytes extra isn't too big of a deal IMO (I mean to use two u32 fields. this won't make not even a slight difference in terms of performance or resource usage), but I am somehow neutral about this, fine either way.

SWIPs/swip-26.md
acud: @mfw78 I'm not sure I fully understand your last message:
> left this method out as there seems to have been an intentional non-committal towards firm typing

by whom? why?

>I am a little concerned that the definition of the type would leak type implementation details into the other. 

define: "other". did not really understand the sentence here.

> Ideally as well, could also compress down to u8 instead of u32, but the idea here was to keep the contents of the frame somewhat within a cache line for a 64 bit machine (to be fair though, I don't have firm benchmarks to support necessitating this).

I'm not 100% sure I understand here. I'm not sure I can in any way assume that anything I write in go can be guaranteed to do a specific number of CPU-Cache ops, not to mention byte alignment etc - we're still going through protobuf which adds so much more abstraction in the middle. If we would be writing this in ASM, OK. Even Rust, Ok - but how can we reason about this without having written the encoders/decoders ourselves?
I think the optimization here is probably not really necessary. It's not like we're doing insane amount of chunk operations here. The whole reserve size is a few million chunks...

acud: By the way, since protobuf uses varint encoding for the chunk types, I am fine with leaving things as they are if this throws a wrench into getting this done. It is fine, and doing enums vs. separate fields is really a nitpick from my side (we have bigger problems to solve and it's probably not worth spilling too many words on).

acud (approved):

significance:
fine either way is the perfect conclusion to every discussion 🧡

significance (approved):

significance:
reference: https://github.com/ethersphere/bee/pull/5445

@zsgsp would be good to get this in the pipeline 🧙‍♂️

PR #69, opened 2025-03-11
add swip-35: negative incentives epic WIP

zelig:
WIP

### DIFF
--- a/SWIPs/swip-35.md
+++ b/SWIPs/swip-35.md
@@ -1,45 +1,90 @@
 ---
 SWIP: 35
-title: Feed Wrapping
+title: Negative Incentives (Epic)
 author: Viktor Trón <viktor@ethswarm.org> (@zelig)
-discussions-to: https://discord.gg/Q6BvSkCv
-dcreated: 2025-03-09
+discussions-to: discord SWIP channel, https://discord.gg/Q6BvSkCv
+status: Draft
+type: Standards Track
+category: Core
+created: 2025-03-09
 ---
 
 # Advanced storage guarantees 
 
 ## Abstract
 
-This SWIP proposes an entire epic about adding further layers of incentives to the basic positive incentives in Swarm. By adopting the SWIP,  users will be offered a wider range of constructions to secure the persistance of chunks and their metadata. 
+This SWIP proposes an entire epic about adding further layers of incentives to the basic positive incentives in Swarm. By adopting the SWIP, swarm will offer a wider range of constructions for its users to secure the persistance of chunks and their metadata and at the same time allow a low barrier of entry service that node operators with extra storage or bzz future providers are able to increase their revenue with.
 
 ## Motivation
 
-The current storage incentive system has a few shortcomings mainly related to the 
-problem that postage batches are difficult to interpret.
+The current storage incentive system has a few shortcomings mainly related to the problem that postage batches are difficult to interpret. This interpretation problem leads to quite a bit of misunderstanding and misguided (also potentially failing) expectations regarding both the stored amount and the storage period. 
 
-The seemingly most natural way to pay for storage is by the amount and time period. Surely the unit price of storage (per chunk per block) already suggests this. However, for most, a deal (through an act of purchase) entails that the counterparty service can be kept to the terms precisely: i.e., you bought a fixed quota (amount of bytes) for a fixed period (duration starting now). Experience shows that any deviation from this expectation comes as a surprise for the user. 
+These issues are coupled with known privacy concerns regarding the shared ownership of chunks that are transparent to 3rd parties. Effectively information about the publisher as well as the chunks constituting the publication leak the system, somewhat overshadowing the otherwise rather strong privacy claims in swarm.
 
-The current primitive of storage compensation, the postage batch falls short on these primarily because it was designed with the simplicity of the intergration into the redistribution system in mind. 
+The latter issue is solved if the stamps on individual chunks are signed by other nodes. We will introduce *stamp mixer service* of sorts in a way that the system will not lose its capacity to prevent doubly sold postage slots.[^3]
 
-The quota encoded in the batch's slots is not neutral: it also serves to force balanced use across neighbourhoods: maxing out the quota for nighbourhood, it implies maxing out the same subquota for each neighbourhood which requires no variance across NH-s; so the effective batch  utilisation rate increases with overall volume, in other words there is a natural overhead for small batches[^1] 
+[^3]: 
+
+On top of all this: if batches are used one per upload session etc, bookkeeping of "what chunk to link with what batch index" is less relevant; in fact, it matters solely in order to save money in case the file is reuploaded.[^4]
+In this SWIP we introduce a scheme in which uploads are not necessarily using a stamp; even though uploaders can purchase storage for arbitrary long periods.
+No per-chunk metadata needs recording and made available in this case, meaning that the end-user case no longer raises portability concerns.
+
+Last, but not least, enforcing payments for each upload does not really make sense: firstly, users had some difficulty to grasp that given the free-tier for bandwidth inentives allowing free downlad and upload with throttled throughput, how come they need to pay for the storage. This problem was most obvious for really short-lived messages that are used in chat or notifications. Since, in these cases, preservation is not meaningful or even actively not desired, paying for postage sounds not only a unnecessary but outright wasteful. Secondly, the fact that expired chunks are not deleted but are put to the *cache*[^5] poses the question how come there is no direct entry to it. 
+This anomaly can be solved if we introduce free uploads, i.e., uploading chunks to Swarm without a postage stamp. 
+
+
+[^5]: so that popular content can bring bandwidth revenue even if no longer available for storage inecntive rewards.
+
+## Solution
+
+### Free uploads to cache
+
+Just like before, the swap protocol offers free (throttled) as well as paid (max throughput) service through cheques, now, upload also offers free (chunks are put to cache and purged if necessary) as well as paid (chunk put to the reserve until stamp expires) service using postage stamps.
+
+
+### Stamp mixing 
+
+#### Incentive to prevent double signing 
+
+Postage batches primarily serve  to enforce the uniform distribution of chunks, and thereby achieve a balanced load of storage across the network. As part of the solution, a user must not double spend, i.e., assign more than one chunk to the same postage slot. Originally, this is guaranteed by the fact that the owner is interested in not being caught since that may lead to the loss of their own data, i.e.,  data that they themselves had uploaded. If a stamp mixer of sorts is introduced, we must make sure that issuers cannot use the same slots: which happens to be the case if all chunks are insured with *negative incentives* (punitive measures).
+
+#### Contractual terms of storage 
+
+The most natural way to pay for storage is by the *amount* and *time period*. Surely, the unit price of storage (per chunk, per block) already suggests this. However, for most, a deal (through an act of purchase) entails that the counterparty service can be kept to the terms *precisely*: i.e., you bought a fixed quota (amount of bytes) for a fixed period (duration starting now, expiring after a fixed period or on a fixed datte). Experience shows that any deviation from this expectation comes as a surprise for the user. 
+
+The current primitive of storage compensation, the postage batch falls short on these. The primary reason is that it was designed with the simplicity in mind pertaining to the intergration into the redistribution system. 
+
+##### Postage batch utilisation
+
+The quota encoded in the batch's slots is not neutral: as per the previous section, it also serves to force balanced use across neighbourhoods: maxing out the subquota for one neighbourhood implies maxing out the same subquota for each neighbourhood which requires no variance across NH-s; so the effective batch  utilisation rate increases with overall volume, in other words there is a natural overhead for small batches[^1] 
 
 [^1]: or more naturally put, there is wholesale bulk discount.
 
-Incomplete utilisation can be counteracted with making sure you can rearrange between neighbourhoods by either [mining]() or having availability assured by a constant influx of chunks (see section below). On the other hand, with the variance limited, the actual discount can also be made explicit in the current system and therefore user expectations of their quota can [bee managed](https://www.overleaf.com/4919957411cgrncysjqrmv#3b42ca).
+Incomplete utilisation can be counteracted with making sure you can rearrange between neighbourhoods by either [mining]() or having availability guaranteed by a constant influx of chunks (see section below). On the other hand, with the variance limited, the actual discount can also be made explicit in the current system and therefore [user expectations of their quota can bee managed](https://www.overleaf.com/4919957411cgrncysjqrmv#3b42ca).
 
-On the other hand, the redistribution  game is a probabilistic outpayment scheme in the sense that instead of paying -- on each game all the contributors for a recurring fixed proportion of used storage we just pay out the entire reward pot for the period to a random winner at every game. This allows for central pricing sensitive to supply and can always automatically discover the optimal price if we assume a liquid provider's market. As a consequence, users do not need to speculate on the future price of storage at all, they can just trust the oracle to discover the rent price of rent: they can provide the compensation for their quota as a "hot wallet" from which the price of rent is automatically deduced. 
+#### Storage period and expiry
 
-However, this also means that undermonetising the batch in the context of increasing rent price may lead to earlier expiry. This can only be counteracted by having more funds on the batch, so users effectively pay for security (upload and disappear experience). After the intended expiry passes, all funds remaining on the batch can be used for more storage service.[^2]
+On the other hand, the redistribution  game is a *probabilistic outpayment scheme* in the sense that instead of paying -- on each game all the contributors for a recurring fixed proportion of used storage we just pay out the entire reward pot for the period to a random winner at every game. This allows for central pricing sensitive to supply and offers automatic and optimal price discovery as long as we assume a liquid market of operators providing storage. As a consequence, users do not need to speculate on the future price of storage at all, they can just trust the oracle to discover the best price of rent. Users provide the compensation for their quota as a "hot wallet" from which the price of rent is automatically deduced. The postage batch is such a construct: it serves as a hot wallet. The expiry is defined as the onset of insolvency, i.e., whenever the balance goes to zero. 
+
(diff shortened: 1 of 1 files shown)


SWIPs/swip-35.md
```diff
@@ -0,0 +1,176 @@
...
+type: Standards Track
+category: Core
+created: 2025-03-09
+---
+
+# Advanced storage guarantees 
+
+## Abstract
+
+This SWIP proposes an entire epic about adding further layers of incentives to the basic positive incentives in Swarm. By adopting the SWIP, swarm will offer a wider range of constructions for its users to secure the persistance of chunks and their metadata and at the same time allow a low barrier of entry service that node operators with extra storage or bzz future providers are able to increase their revenue with.
```
significance, line 16:
```suggestion
This SWIP proposes an entire epic about adding further layers of incentives to the basic positive incentives in Swarm. By adopting the SWIP, swarm will offer a wider range of constructions for its users to secure the persistence of chunks and their metadata and at the same time allow a low barrier of entry service that node operators with extra storage or bzz future providers are able to increase their revenue with.
```

SWIPs/swip-35.md
```diff
@@ -0,0 +1,176 @@
...
+
+## Motivation
+
+The current storage incentive system has a few shortcomings mainly related to the problem that postage batches are difficult to interpret. This interpretation problem leads to quite a bit of misunderstanding and misguided (also potentially failing) expectations regarding both the stored amount and the storage period. 
+
+These issues are coupled with known privacy concerns regarding the shared ownership of chunks that are transparent to 3rd parties. Effectively information about the publisher as well as the chunks constituting the publication leak the system, somewhat overshadowing the otherwise rather strong privacy claims in swarm.
+
+The latter issue is solved if the stamps on individual chunks are signed by other nodes. We will introduce *stamp mixer service* of sorts in a way that the system will not lose its capacity to prevent doubly sold postage slots.
```
significance, line 24:
```suggestion
The latter issue is solved if the stamps on individual chunks are signed by other nodes. We will introduce a *stamp mixer service* of sorts in a way that the system will not lose its capacity to prevent doubly sold postage slots.
```

SWIPs/swip-35.md
```diff
@@ -0,0 +1,176 @@
...
+In this SWIP we introduce a scheme in which uploads are not necessarily using a stamp; even though uploaders can purchase storage for arbitrary long periods.
+No per-chunk metadata needs recording and made available in this case, meaning that the end-user case no longer raises portability concerns.
+
+Last, but not least, enforcing payments for each upload does not really make sense: firstly, users had some difficulty to grasp that given the free-tier for bandwidth inentives allowing free downlad and upload with throttled throughput, how come they need to pay for the storage. This problem was most obvious for really short-lived messages that are used in chat or notifications. Since, in these cases, preservation is not meaningful or even actively not desired, paying for postage sounds not only a unnecessary but outright wasteful. Secondly, the fact that expired chunks are not deleted but are put to the *cache*[^5] poses the question how come there is no direct entry to it.
```
significance, line 30:
```suggestion
Last, but not least, enforcing payments for each upload does not really make sense: firstly, users had some difficulty to grasp that given the free-tier for bandwidth inentives allowing free download and upload with throttled throughput, how come they need to pay for the storage. This problem was most obvious for really short-lived messages that are used in chat or notifications. Since, in these cases, preservation is not meaningful or even actively not desired, paying for postage sounds not only a unnecessary but outright wasteful. Secondly, the fact that expired chunks are not deleted but are put to the *cache*[^5] poses the question how come there is no direct entry to it. 
```

SWIPs/swip-35.md
```diff
@@ -0,0 +1,176 @@
...
+
+#### Contractual terms of storage 
+
+The most natural way to pay for storage is by the *amount* and *time period*. Surely, the unit price of storage (per chunk, per block) already suggests this. However, for most, a deal (through an act of purchase) entails that the counterparty service can be kept to the terms *precisely*: i.e., you bought a fixed quota (amount of bytes) for a fixed period (duration starting now, expiring after a fixed period or on a fixed datte). Experience shows that any deviation from this expectation comes as a surprise for the user.
```
significance, line 53:
```suggestion
The most natural way to pay for storage is by the *amount* and *time period*. Surely, the unit price of storage (per chunk, per block) already suggests this. However, for most, a deal (through an act of purchase) entails that the counterparty service can be kept to the terms *precisely*: i.e., you bought a fixed quota (amount of bytes) for a fixed period (duration starting now, expiring after a fixed period or on a fixed date). Experience shows that any deviation from this expectation comes as a surprise for the user. 
```

SWIPs/swip-35.md
```diff
@@ -0,0 +1,176 @@
...
+
+#### Contractual terms of storage 
+
+The most natural way to pay for storage is by the *amount* and *time period*. Surely, the unit price of storage (per chunk, per block) already suggests this. However, for most, a deal (through an act of purchase) entails that the counterparty service can be kept to the terms *precisely*: i.e., you bought a fixed quota (amount of bytes) for a fixed period (duration starting now, expiring after a fixed period or on a fixed datte). Experience shows that any deviation from this expectation comes as a surprise for the user. 
+
+The current primitive of storage compensation, the postage batch falls short on these. The primary reason is that it was designed with the simplicity in mind pertaining to the intergration into the redistribution system.
```
significance, line 55:
```suggestion
The current primitive of storage compensation, the postage batch falls short on these. The primary reason is that it was designed with the simplicity in mind pertaining to the integration into the redistribution system. 
```

SWIPs/swip-35.md
```diff
@@ -0,0 +1,176 @@
...
+
+Let `s` and `e` be the start and end data of storage, period is `e-s` in units used in the oracle price. Let us propose the following scenario: owner `A` buys a postage batch `B` for `2^C` chunks and escrows amount `D` onto `B`. Now if we assume that price of storage per chunk per time unit at `s` is `p_s`, then we can guess that the expiry is `e'<=e` iff `D/(2^C*p_s)` then A assumes price will decrease. However, if the average price increases, then the effective period ends d units sooner if d is the smallest number such that the average price over the period `[s, e]` the cumulative rent goes above D `\Sum_{i=s}^e p_i*2^C >= D.`
```
significance, line 69:
```suggestion
Let `s` and `e` be the start and end data of storage, period is `e-s` in units used in the oracle price. Let us propose the following scenario: owner `A` buys a postage batch `B` for `2^C` chunks and escrows amount `D` onto `B`. Now if we assume that price of storage per chunk per time unit at `s` is `p_s`, then we can guess that the expiry is `e'<=e` iff `D/(2^C*p_s)` then `A` assumes price will decrease. However, if the average price increases, then the effective period ends d units sooner if d is the smallest number such that the average price over the period `[s, e]` the cumulative rent goes above D `\Sum_{i=s}^e p_i*2^C >= D.`
```

SWIPs/swip-35.md
```diff
@@ -0,0 +1,176 @@
...
+
+#### Upload to the reserve, filling batches
+
+Insurers can collect and then publish chunks with their own stamp. The batch that they use will be grouping chunks by
+
+1. H(address); locality of their node address
+2. end; datetime of expiry of storage
+3. start; arrival time of request 
+
+As long as 3. is undifferentiating due to a large number of (near) simultaneous requests
```
significance, line 120:
```suggestion
As long as 3 is undifferentiating due to a large number of (near) simultaneous requests
```

SWIPs/swip-35.md
```diff
@@ -0,0 +1,176 @@
...
+1 and 2 will not help align identity to chunk or even chunk to chunk easily. This is why push-syncing all the chunks in the batch can be delayed: exactly in order to achieve that a larger number of simultaneous requests get published together. Ideally, the time of delay coincides with the time of batch completion, i.e., whenever the batch (prepared to support the time until original uploaders want the chunk to be stored) is fully utilised, it is closed and all the chunks are uploaded. The problem is that if the traffic is not great, especially long term storage (more expensive), publishing may need to wait beyond the chunks of the file already disappear from the swarm's cache. These chunks therefore need to be first uploaded with a rather short period just to survive and later maybe also  have their subscription renewed. Such a strategy creates more usage for the short periods which then leads to the ability to fill bigger batches faster and more efficiently: the bigger the batch, the bigger the anonimity set, thus, the better the anonimity that you get.
```
significance, line 121:
```suggestion
1 and 2 will not help align identity to chunk or even chunk to chunk easily. This is why push-syncing all the chunks in the batch can be delayed: exactly in order to achieve that a larger number of simultaneous requests get published together. Ideally, the time of delay coincides with the time of batch completion, i.e., whenever the batch (prepared to support the time until original uploaders want the chunk to be stored) is fully utilised, it is closed and all the chunks are uploaded. The problem is that if the volume of traffic is not great, especially traffic relating to long term storage (more expensive), publishing may be delayed until a time after which the chunks of the file have already disappeared from the swarm's cache. These chunks therefore need to be first uploaded with a rather short period just to survive and later maybe also have their subscription renewed. Such a strategy creates more usage for the short periods which then leads to the ability to fill bigger batches faster and more efficiently: the bigger the batch, the bigger the anonymity set, thus, the better the anonymity that you get. 
```

SWIPs/swip-35.md
```diff
@@ -0,0 +1,176 @@
...
+In fact, this opens up the possibility that providers keep alive batches practically for short time periods representing a unit period that the others are multiples of. In order to know how long the relevant chunks are meant to be stored,
+nodes save information about the original request.
+
+
+
+## Specification
+
+## Evaluation
+
+Even though treating the postage batch as a hot wallet has quite a few advantages, such an insurance scheme bring improvement in several areas:
```
significance, line 132:
```suggestion
Even though treating the postage batch as a hot wallet has quite a few advantages, such an insurance scheme brings improvements in several areas:
```

SWIPs/swip-35.md
```diff
@@ -0,0 +1,176 @@
...
+
+  - **privacy**:
+    - effective *stamp mixing*
+    - complete obfuscation (and plausible deniability) of chunk originator
+    - no leakage of shared ownership through signature by the same batch owner
+  - **precision**:
+    - exact period
+    - exacy amount
```
significance, line 140:
```suggestion
    - exact byte amount 
```

SWIPs/swip-35.md
```diff
@@ -0,0 +1,176 @@
...
+  	- free upload into cache
+    - easier to understand pricing
+    - potentially complete utilisation of batches leading to cheaper prices or easier to coordinate subsidies 
+    - prepurchase (keeping batches open) makes sense (speculation on price deal)
+    - moderate future price speculation needed
+    - 'pay once, store forever' now available to providers (only)
+    - no benefit from rent price drop
+  - **portability**:
+    - only providers need to keep track of batches and batch utilisation
+    - since users got all they need from just knowing that a file/root was insured/kept alive
```
significance, line 156:
```suggestion
    - users get all the assurance they need from just knowing that a file/root was insured/kept alive
```

SWIPs/swip-35.md
```diff
@@ -0,0 +1,176 @@
...
+
+## Implementation notes and roadmap
+
+Due to the complexity of this epic, components will be  discussed in separate SWIPs:
+- free postage
+- global pinning, incentives, notifications
+- stamp mixing, requests, receipts and address indexes
+- challanges with inclusion proofs etc
```
significance, line 164:
```suggestion
- challenges with inclusion proofs etc
```

SWIPs/swip-35.md
```diff
@@ -0,0 +1,325 @@
...
+#### Free uplods to cache
+
+Messages of real-time chat or notifications or frequently updated indexes without history are just obvious examples of data that are meant to be short-lived. Since, in these cases, preservation beyond temporary caching is unnecessary, paying for postage is wasteful. 
+Such content meant not to be stored should just be be uploadable without a postage stamp and put to the cache and not to the reserve. In order to mitigate targetted attacks, bandwidth compensation must be adapted to express as well as transfer the cost between uploader and  storer in case postage stamps  are not able to.
+
+## Solution
+
+### Negative incentives
+The lack of individual accountability makes Swarm's storage incentivisation limited as a security measure against data loss. Introducing non-compensatory insurance, on the other hand, adds an additional layer of *negative incentives*. In other words, the threat of punitive measures (losing the stake deposited against foul play) is introduced
+in order to compell storage providers to ensure reliability of data retention and/or prompt retrievability for users.
```
nugaon, line 79: Data retrievability is not yet outlined in the spec (just for note since we touched upon it before)

SWIPs/swip-35.md
```diff
@@ -0,0 +1,325 @@
...
+
+Let `s` and `e` be the start and end date of storage, period is `e-s` in units used in the oracle price. Let us propose the following scenario: owner `A` buys a postage batch `B` for `2^C` chunks and escrows amount `D` onto `B`. Now if we assume that price of storage per chunk per time unit at `s` is `p_s`, then we can guess that the expiry is `e'<=e` iff `D/(2^C*p_s)` then A assumes price will decrease. However, if the average price increases, then the effective period ends d units sooner; if d is the smallest number such that the average price over the period `[s, e]` the cumulative rent goes above D `\Sum_{i=s}^e p_i*2^C >= D.`
+
+In other words, increasing rent price may lead to earlier expiry. This can only be counteracted by having more funds on the batch, so users effectively pay for security (*upload and disappear experience*). If the intended expiry passes without the batch running out of funds, then storage just continues, or alternatively all funds remaining on the batch can be used for more storage service.[^12]
+
+[^12]: Using the current balance to create another batch (potenitally different size, period and owner) is currently not supported but it should be.
```
nugaon, line 142: what would be the upside? aren't the topup/dilute functions sort out such needs?

SWIPs/swip-35.md
```diff
@@ -0,0 +1,325 @@
...
+
+If insurance providers themselves are to secure the quota using their own postage batch, then the surplus is all the more likely to be filled. Completely filled batches result in a lower unit prices of storage. 
+
+The stumbler construction is virtually a stamp mixer in that it blinds any originator address's association with chunks and also obfuscates their co-constituency (belonging together of a set of chunks).
+Since association by postage batch was always the wierd anomaly, solving it is a major improvement in swarm's privacy.
+
+
+#### Distribution of responsibility: providers in the overlay address space
+  
+A liquid quota market is best provided by a *decentralised service network*: served by swarm nodes themselves. Ideally providers will have a density not too dissimilar to the nodes providing the reserve, but the system also needs to be prepared to provide service in the case of a lot (a couple of orders of magnitude) fewer or more nodes. This is taken care of by the swear contract, which maintains a balanced set of providers according to *smart neighbourhood management* ([described in a separate SWIP (SWIP-39)](https://github.com/ethersphere/SWIPs/pull/74))
```
nugaon, line 171: in case of insurers, it is unknown what would be the "garbage collection" of not participating actors.

SWIPs/swip-35.md
```diff
@@ -0,0 +1,325 @@
...
+
+A *withholding attack* is when an insurer is grieving[^7] consumers by only providing the chunk if challenged. If a chunk is to be retrievable from a different neighbourhood (belong to the reserve based on the chunk's address) than the one that insures it, then the onus of retrievability falls on nodes that are independent in terms of operation.
+
+[^7]: a grieving attack is where the adversary is able to make their victim(s) suffer losses, time or other inconvenience without them gaining.
+
+We propose that chunks have a *secondary address*, defined as the (legacy Keccak) hash of the primary chunk address with or without the first bit flipped (bitwise XOR the first byte with 0x80). This will place roughly one in every 256 chunk into the opposite side of the network which still keeps the balance of storage.
+
+So when nodes swear on to provide insurance (or insurance/availability level X service), they register their address with a stake on a (the?) staking contract.  It is important that providers should be distributed evenly in neighbourhoods, ideally of the same size as that dictated by the reserve.
```
nugaon, line 179: the proposal does not detail the stake prices for running an insurance service. Is it the same all across the network? Does it influence in any way how many chunks they can stamp? Does it vary over time?

SWIPs/swip-35.md
```diff
@@ -0,0 +1,325 @@
...
+They can keep the chunk in this batch retrieveable by regularly topping up the batch if needed. Since the hashes are uniformly distributed, this batch is expected to be naturally balanced as it should be. So it is quite realistic for the node to be able to fill it. 
+The request will reference a specified expiry and the batch needs to have a balance at least until that date. 
+
+
+### Insurance
+
+#### Punitive measures
+
+Unlike in the case of bandwidth incentives where retrievals are immediately accounted and settled, long-term storage guarantees are promissory in nature and do not allow the instant gratification of forwarding/backwarding. It is only after the end of its validity that one can check if a promise has been kept.
+Nodes wanting to sell storage space should have a stake verified and locked-in at the time of making their promise. This implies that nodes need to register in advance, agreeing to the contract by placing a security deposit. Once registered, a node may start selling storage promises. While their registration is active, if they are found to have lost a chunk that was covered by their promise, they stand to lose their deposit.
```
nugaon, line 192: There is also a service reliability concern: what happens with the insured chunks of different users if the provider got slashed? They cannot challange the provider anymore since there is no stake to lose. It implies that one provider slash affects multiple users in the same time.

SWIPs/swip-35.md
```diff
@@ -0,0 +1,325 @@
...
+- *SWAP*: The swarm accounting protocol makes sure that peers exchange chunks
+and receipts, triggering swap accounting and settlement. This must be used for payment 
+in order for true uploaders to remain anonymous.
+- *SWEAR*: Storer nodes register with their stake and make an implicit promise to store the chunks whose secondary address falls within their area of responsibility.
+- *SWINDLE* -- Storers can be challanged on their promise through an on-chain litigation process.
+
+In case an escrow is used, the reward is vested over a period of 
+the storer is only awarded the locked funds if they are able
+to provide *proof of entitlement*. This procedure is analogous to the postage
+stamp redistribution game. Out-payments conditional on proofs of custody can be implemented in a way similar to the reserve redistribution game.
```
nugaon, line 213: if there are multiple beneficiaries, the users can overissue cheques over their deposited balance anytime, so maybe providers could have the option to drop commitment in that case (similarly closing a connection/blacklisting in the case of bandwidth incentives).

SWIPs/swip-35.md
```diff
@@ -0,0 +1,325 @@
...
+
+The SWEAR contract allows nodes to register their public key to become
+accountable participants in the swarm. Registration involves sending
+the deposit to the SWEAR contract, which serves as collateral in case
+the terms that registered nodes "swear" to keep are violated (i.e. nodes
+do not keep their promise to store). The registration is open-ended  Users of Swarm should be able to count on the loss of deposit as
+a disincentive against foul play for as long as the insurer node has open commitments.
+Therefore, even a well-performing Swarm node is only eligible to reclaim their
+deposit after all their commitments are closed or taken over by new nodes.
```
nugaon, line 222: how is it verifiable? These are off-chain commitments... 
my first shoot is that to have a withdrawal challange period when insurer issues a reclaim requests with a state root hash and anyone can provide a fraud proof against it.

But what is the structure and logic between commitment state versions of the provider? Because a record of a commitment should appear in every version of commitment state versions until its lifetime.

If all previous approach is correct, then this system needs active watchtowers/verifier actors scanning for fraudalence between every epoch... and that is not really a well incentivized or accountable role.

PR #76, opened 2025-08-11
SWIP-40: Standard for recursive encrypted chunk reading

tmm360:
New SWIP proposal to add compatibility with recursively encrypted chunks reading.

### DIFF
--- a/SWIPs/swip-40.md
+++ b/SWIPs/swip-40.md
@@ -13,21 +13,21 @@ created: 2025-08-11
 
 ## Simple Summary
 <!--"If you can't explain it simply, you don't understand it well enough." Provide a simplified and layman-accessible explanation of the SWIP.-->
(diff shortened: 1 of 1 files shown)


tmm360:
Reviewing the encryption procedure, emerged that XOR doesn't permit to encrypt chunk's span. This could help to scan the key, deriving original span from the chunk length, if it is a leaf. Probably a better chunk encryption method must be adopted, to allow also span's encryption.

tmm360:
Withdrawing the proposal because I've discovered that I can make my implementation compatible with the chunks encryption algorithm already implemented.

significance:
hi @tmm360 came here to respond and found it closed! must admit to being a little confused as to what you are trying to achieve, but interested to find out!

I believe EncryptChunk does indeed encrypt the span and does (should!) not leak information about the key, and the keys can be random as you please so that you may distribute the chunks as you please.

perhaps we can discuss more in discord, feel free to tag me @sigsigsig 😌

PR #80, opened 2025-11-08
Add SWIP-42: Mined Owner Chunks (MOCs) for efficient notifications

zelig:


zelig:
@copilot provide description

significance:
very interesting. it would indeed be very good. could the current protocol support the polling? 

i foresee maybe scaling issues as service nodes would have to maintain many more connections to consumer nodes, perhaps something that could be mitigated by providing nodes with incentive to do so and some balancing defacto by per-node limits or otherwise more specific

maybe one for the good problems to have bucket that, though. bring it on! ( :

zelig:
Well, }I dont think any extra connections will be needed at all. Instead nodes will try to use any of their already established connections (to create a MIC/MOC address) neighbourhood

zelig:
dependencies: 
pubsub broker discovery described in [pubsub epic](https://github.com/ethersphere/SWIPs/pull/93)
=> depends on MEX in  [SWIP-59](https://github.com/ethersphere/SWIPs/pull/103)
=> depends on MIC/MOC [SWIP-42](https://github.com/ethersphere/SWIPs/pull/80)

PR #83, opened 2026-01-16
Swip 44 remove skipped rounds guard

significance:
This SWIP proposes removing the skipped rounds guard check from the redistribution contract.


significance:
after some discussion it was understood this was in fact guarding against other angles of attack

PR #84, opened 2026-01-21
SWIP-45: Fork Digest and BzzAddress Encoding

mfw78:
## Summary

This SWIP introduces a fork digest mechanism for network upgrade coordination and cleans up the BzzAddress underlay encoding by replacing custom serialization with idiomatic protobuf.

### Fork Digest
- **4-byte identifier** derived from genesis hash and current fork version
- **Handshake integration** - nodes reject connections where fork digests do not match
- **Hive protocol integration** - fork-aware peer gossip filtering prevents accumulating stale peer lists

### BzzAddress Encoding
- **`repeated bytes underlays`** replaces `bytes underlay` with custom encoding
- **Removes custom serialization** - no more `0x99` magic prefix or varint length encoding
- **Protobuf handles wire format** - separates wire encoding from application logic

### Signature Scheme (v1)
- **Simple concatenation** of multiaddr bytes for signature (no padding needed)
- **Removes redundant `"bee-handshake-"` prefix** (EIP-191 already provides domain separation)
- **Includes fork digest** to bind signatures to specific forks

## Migration Path

1. **Release N** - Fork digest optional, accepts both legacy and new formats
2. **Release N+1** - Only new format accepted, legacy rejected

## Feedback Requested

The exact composition of the genesis hash requires further discussion. Candidates for inclusion:

- `network_id` (required)
- `genesis_timestamp`
- Contract addresses (postage stamp, staking, redistribution)
- Other network-specific parameters

### DIFF
--- a/SWIPs/swip-fork-digest.md
+++ b/SWIPs/swip-fork-digest.md
@@ -1,0 +1,185 @@
+---
+swip: <to be assigned>
+title: Fork Digest for Network Upgrade Coordination
+author: mfw78 (@mfw78)
+status: Draft
+type: Standards Track
+category: Networking
+created: 2026-01-21
+---
+
+## Simple Summary
+
+Introduce a fork digest mechanism that enables nodes to identify compatible peers during network upgrades via the handshake and Hive peer gossip.
+
+## Abstract
+
+A fork digest is a 4-byte identifier derived from the network configuration and current fork version. By including the fork digest in both the handshake protocol and Hive peer advertisements, nodes can efficiently filter incompatible peers before connection and avoid gossiping peers that recipients cannot use.
+
+## Motivation
+
+Swarm lacks a standardised mechanism for:
+
+1. **Peer compatibility detection.** Nodes cannot determine protocol compatibility before establishing connections.
+2. **Efficient peer gossip.** Hive broadcasts all known peers regardless of fork compatibility, wasting bandwidth.
+3. **Graceful upgrades.** During network upgrades, incompatible nodes waste resources attempting failed connections.
+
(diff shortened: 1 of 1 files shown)


significance:
@mfw78 thank you for this! 

i feel like there are two issues here. i think it would be good to separate them for clarity of discussion 😌

significance:
btw i hereby object to the nomenclature "fork" being used here. this is not a fork, it is just a protocol version migration

a fork is something else 🍽️ 

edit: 

for clarity, other things a fork is:

  1. foss fork — someone takes the code and goes their own way, usually because the community split (libreoffice from openoffice)
  2. github fork — just clicking a button to copy a repo so you can make a PR, no drama involved
  3. soft fork — backwards-compatible blockchain rule change, tightens the rules, no chain split
  4. hard fork (contentious) — backwards-incompatible change where people disagree and the chain actually splits in two, both sides survive (ethereum classic)
  5. hard fork (routine) — same mechanism but everyone agrees, old chain just dies, it's basically a network upgrade
  6. protocol/defi fork — copy-pasting someone's smart contracts and deploying them with a new name (sushiswap from uniswap)
  7. a novel item of cutlery denounced in 1004 by the Catholic church as an excessive vanity and an affront to god

mfw78:
> @mfw78 thank you for this!
> 
> i feel like there are two issues here. i think it would be good to separate them for clarity of discussion 😌

If you perceive that there are two issues here, can you _please_ explicitly explain how you see the two issues being divided.

significance:
1. BzzAddress encoding and sig scheme
2. Protocol version migration process

PR #90, opened 2026-04-16
SWIP 191 & SWIP 192

lat-murmeldjur:
In these two SWIPs I propose a more resilient decentralised commenting engine for swarm.

Combined, the conventions proposed here can provide a substantially more censorship-resistant decentralised commenting system that can operate entirely on Swarm. Besides commenting, the anythread engine can serve a number of social purposes, such as creating searchable user registries where anyone can sign up, sending friend requests without previously established communications channels to discovered users, and publishing and discovering new content.

### DIFF
--- a/SWIPs/swip-191-192.md
+++ b/SWIPs/swip-191-192.md
@@ -1,0 +1,276 @@
+---
+swip: 191
+title: Efficient multiple-version SOC exhaustive lookup
+author: lat-murmeldjur
+status: Draft
+type: Standards Track
+category: Primitives
+created: 2026-04-16
+---
+
+---
+swip: 192
+title: Censorship-resistant decentralised commenting on Swarm
+author: lat-murmeldjur
+status: Draft
+type: Standards Track
+category: Primitives
+created: 2026-04-16
+---
+
+
+# SWIP 191 - Efficient multiple-version SOC exhaustive lookup
+# SWIP 192 - Censorship-resistant decentralised commenting on Swarm
+
+## Problem statement
+
+Single Owner Chunks complement content-addressed chunks by allowing a fixed address, derived from an owner and a topic, to point to arbitrary dynamic content. This makes it possible to create a deterministically addressed root chunk whose address can be calculated by anyone who knows the SOC owner and topic. Sequential feeds are built on top of this primitive: feed updates are consecutive SOCs whose addresses can be calculated by anyone who knows the feed owner and the feed topic, with the topic for each update derived from the feed topic and the update index itself.
+
+The anythread design shows that this mechanism can be repurposed into an open commenting and threading system for arbitrary subjects. A thread is defined by a topic string. The shared feed owner is derived from that topic by using it as a publicly known private key seed in such a way that anyone can write with it, and the feed topic is likewise derived from the same topic string. This creates a feed associated to any subject, such as a Swarm reference, a web2 URI or a label or statement, and that feed can be written by anyone.
+
+To keep comments attributable to the actual writer, anythread introduces one level of indirection. The shared feed does not primarily store comment bodies. Instead, when a participant has a comment for a topic, they publish their own private feed owner identity on the shared anythread feed. The actual comment then resides on a separate feed for the same topic, but owned by that participant alone. This keeps authorship clear while avoiding the need for the shared feed itself to be trusted as the final storage location of the comment content.
+
+This construction also exposes an emergent property of SOCs. If multiple actors use a common SOC owner, they may upload different contents under the same SOC address, including concurrently. If this is not handled explicitly, different parts of a neighborhood in the network may observe different contents for the same SOC address.
+
+The Graffiti Single/Several Owner Chunk (GSOC) primitive addresses this by making distinct versions of the same SOC visible within the neighbourhood through pull sync, even when they were uploaded with different postage stamps. This avoids the simple partitioning scenario, but it introduces a new requirement: once multiple valid contents may exist for the same SOC address, applications need a practical and reliable way to retrieve all versions of that SOC.
+
+The purpose of SWIP 191 is to provide the retrieval-side mechanism required for efficient exhaustive lookup of multiple versions of a single SOC. 
+The purpose of SWIP 192 is to make anythread significantly more resistant to censorship and operational failure modes.
+
+## SWIP 191 - Efficient multiple-version SOC exhaustive lookup
+
+### Motivation
+
+If multiple valid versions of the same SOC may exist, a plain retrieval request is no longer enough for an application that needs completeness. A node may return one valid version, but the caller still does not know whether other valid versions exist. What is required is a way to constrain retrieval requests so that the caller can systematically enumerate the entire set of wrapped content addresses associated with a given SOC.
+
+The goal is not to change the integrity model of Swarm. Returned chunks must still validate against the requested SOC. The goal is to extend retrieval so that an application can ask for a version matching additional conditions and, by repeating such requests deterministically, discover every version that exists.
+
+### Retrieve protocol extensions
+
+Two additional query fields are sufficient for this purpose.
+
+1. `prefix`
+
+   This field is conceptually a bit prefix. At the protobuf level it can be represented as bytes, with the understanding that the query specifies a precise prefix length in bits rather than only in whole bytes. The condition is:
+
+   Return a version of SOC `R` whose wrapped content address begins with the bits specified by `prefix`.
+
+2. `other_than`
+
+   This field is a 32-byte wrapped content address. The condition is:
+
+   Return a version of SOC `R` whose wrapped content address is different from the address specified in `other_than`.
+
+If both fields are present, both conditions must be held. In other words, the request means:
+
+Return a version of SOC `R` whose wrapped content address matches `prefix` and is different from `other_than`.
+
+These additions imply only a small set of changes at the node level:
+
+- the two fields must be added to the retrieve protocol;
+- response validation must verify that the returned chunk satisfies the requested conditions if the returned chunk is an SOC;
+- chunk-store lookup logic must be able to select a matching version when several versions of the same SOC exist.
+
+This is sufficient to support efficient exhaustive lookup without changing the basic validity rules for SOC retrieval.
+
+### Exhaustive lookup at an application level
+
+An external application (for example a dapp, or the anythread library) can use the two fields above to enumerate all versions of a target SOC with a reasonably low number of requests. The key observation is that wrapped content addresses can be explored as ranges of binary prefixes. Each request either discovers a new address in a range or proves that no additional address exists in that range under the given exclusion condition.
+
+A compact example makes the idea easier to follow.
+
+Assume the target SOC address is `R`.
+
+1. Retrieve `R` without any additional condition.
+   - Suppose the response contains wrapped content address `A`, and `A` begins with `0000...`.
+
+2. Retrieve `R` again with `other_than = A`.
+   - If this returns nothing, then `A` is the only known version and the lookup is complete.
+   - Otherwise suppose the response contains a second address `B`, and `B` begins with `0001...`.
+
+3. At this point `A` and `B` share the prefix `000`. This tells us three things:
+   - there may still be further versions under the `0000...` branch besides `A`;
+   - there may still be further versions under the `0001...` branch besides `B`;
+   - there may also be versions in the neighbouring ranges `001...`, `01...`, and `1...`, which are not yet covered by either known address.
+
+   The next requests are therefore:
+
+   - `prefix = 0000`, `other_than = A`
+   - `prefix = 0001`, `other_than = B`
+   - `prefix = 001`
+   - `prefix = 01`
+   - `prefix = 1`
+
+Each request has one of two outcomes. It either returns a new wrapped content address, or it proves that the queried range contains no additional version satisfying the request. By repeating this process only for unexplored ranges, the application can cover the full address space without issuing redundant requests.
+
+The opposite case is also illustrative. If the first two retrieved addresses were `C = 1...` and `D = 0...`, then they would already diverge at the first bit. In that case the next requests would be:
+
+- `prefix = 1`, `other_than = C`
+- `prefix = 0`, `other_than = D`
+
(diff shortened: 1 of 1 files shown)


SWIPs/swip-191-192.md
```diff
@@ -0,0 +1,266 @@
...
+author: lat-murmeldjur
+status: Draft
+type: Standards Track
+category: Primitives
+created: 2026-04-16
+---
+
+
+# SWIP 191 - Efficient multiple-version SOC exhaustive lookup
+# SWIP 192 - Censorship-resistant decentralised commenting on Swarm
```
nugaon, line 13: this SWIP title comes after this section.

SWIPs/swip-191-192.md
```diff
@@ -0,0 +1,266 @@
...
+
+A property of the algorithm is that it avoids dispatching multiple requests for the same address-space range. It only explores branches that are known to matter and only tests gaps that are not already covered.
+
+## SWIP 192 - Censorship-resistant decentralised commenting on Swarm
+
+### What SWIP 191 solves for anythread
+
+SWIP 191 provides a deterministic way to discover all versions of a given SOC with reasonable efficiency. For anythread this is directly useful. If an attacker attempts to censor a specific feed index by publishing additional versions of the same SOC, retrieval no longer has to rely on chance. The application can enumerate all versions of that index and recover the intended one.
+
+This does not solve spam or flooding. Anythread is meant to be censorship-resistant, and content moderation is outside the scope of this proposal. The purpose here is narrower: to make it harder for a participant to hide legitimate thread updates by exploiting the fact that a shared feed owner allows many writers to target the same SOC address.
```
nugaon, line 176: flood is exponentially more expensive with GSOC, since all updates fall into the same neighborhood area and for writing you need to buy a postage batch with which you must purchase the same store counts in each neighborhood, not only your targeted one.

SWIPs/swip-191-192.md
```diff
@@ -0,0 +1,266 @@
...
+
+This construction also exposes an emergent property of SOCs. If multiple actors use a common SOC owner, they may upload different contents under the same SOC address, including concurrently. If this is not handled explicitly, different parts of a neighborhood in the network may observe different contents for the same SOC address.
+
+The Graffiti Single/Several Owner Chunk (GSOC) primitive addresses this by making distinct versions of the same SOC visible within the neighbourhood through pull sync, even when they were uploaded with different postage stamps. This avoids the simple partitioning scenario, but it introduces a new requirement: once multiple valid contents may exist for the same SOC address, applications need a practical and reliable way to retrieve all versions of that SOC.
+
+The purpose of SWIP 191 is to provide the retrieval-side mechanism required for efficient exhaustive lookup of multiple versions of a single SOC. 
+The purpose of SWIP 192 is to make anythread significantly more resistant to censorship and operational failure modes.
+
+## SWIP 191 - Efficient multiple-version SOC exhaustive lookup
```
nugaon, line 30: `other_than` is a nice extension of the constrained SOC request format,
the serialization could happen in the same manner as is detailed in the [GSOC proposal](https://github.com/nugaon/SWIPs/blob/graffiti-soc/SWIPs/swip-draft_graffiti-soc.md#change-in-the-chunk-retrieval-protocol)
so optionType `0` could be the prefixed way to query and `1` could be the `other_than`.

SWIPs/swip-191-192.md
```diff
@@ -0,0 +1,266 @@
...
+The security parameter works by narrowing the accepted part of the wrapped-content address space.
+
+At security parameter `p = 0`, the application discovers any valid version of the SOC.
+
+At security parameter `p > 0`, the application only browses and writes versions whose wrapped content address matches a designated `p`-bit prefix. A natural choice is to require the wrapped content address to match the first `p` bits of the SOC address itself. Writers can satisfy this condition by introducing a salt after the commenting-identity portion of the SOC payload and varying that salt until the wrapped content address falls into the accepted prefix range.
+
+This has two useful effects:
+
+- it raises the amount of work required to generate a version that the application will treat as relevant;
+- it gives the reader a narrower starting range for exhaustive lookup.
```
nugaon, line 197: it has strong impact on the UX, increasing load time and resource usage for honest clients, who need to do the same work as flooders.

SWIPs/swip-191-192.md
```diff
@@ -0,0 +1,266 @@
...
+
+The pager feed is a second feed associated with the same thread and topic. Its purpose is not to replace the base feed, but to summarise occupancy over larger index ranges.
+
+The pager feed works as follows:
+
+- a pager size is chosen at application level, for example `32`;
+- when a writer publishes base-feed index `N`, it also publishes the same content at pager-feed index `floor(N / pager_size)`;
+- each pager-feed index therefore corresponds to a `pager_size` long section of the base-feed indexes;
+- because many base indexes map to the same pager index, a single pager index may legitimately accumulate multiple versions;
+- the writer is expected to publish the base entry and the pager entry with the same postage stamp, probably making them expire together.
```
nugaon, line 229: with the same postage batch, they should use different stamps

SWIPs/swip-191-192.md
```diff
@@ -0,0 +1,266 @@
...
+
+This gives the reader an efficient look-ahead mechanism. If the pager size is `32`, then checking the next `16` pager indexes tells the reader whether any updates exist in the next `512` base-feed indexes. Since SWIP 191 already provides a method to enumerate all versions of a pager index, the application can recover the pager information efficiently even though each pager index may hold multiple versions.
```
nugaon, line 231: this does not give any guarantee; all records can be garbage collected on a pager feed address over time.
thereby, the same corrections will be required to fetch records as it was a simple base feed

SWIPs/swip-191-192.md
```diff
@@ -0,0 +1,266 @@
...
+
+The base feed and the pager feed serve different purposes and are both useful:
+
+- the base feed spreads writes across more neighbourhoods, which makes censorship by local manipulation more difficult;
+- the strong correspondence between base indexes and pager indexes helps writers locate where to write in the pager feed without concentrating too many writes onto a single overloaded SOC by accident;
```
nugaon, line 236: too many writes on a SOC by accident? this point is not really clear.

SWIPs/swip-191-192.md
```diff
@@ -0,0 +1,266 @@
...
+- when a writer publishes base-feed index `N`, it also publishes the same content at pager-feed index `floor(N / pager_size)`;
+- each pager-feed index therefore corresponds to a `pager_size` long section of the base-feed indexes;
+- because many base indexes map to the same pager index, a single pager index may legitimately accumulate multiple versions;
+- the writer is expected to publish the base entry and the pager entry with the same postage stamp, probably making them expire together.
+
+This gives the reader an efficient look-ahead mechanism. If the pager size is `32`, then checking the next `16` pager indexes tells the reader whether any updates exist in the next `512` base-feed indexes. Since SWIP 191 already provides a method to enumerate all versions of a pager index, the application can recover the pager information efficiently even though each pager index may hold multiple versions.
+
+The base feed and the pager feed serve different purposes and are both useful:
+
+- the base feed spreads writes across more neighbourhoods, which makes censorship by local manipulation more difficult;
```
nugaon, line 235: what if the hostile actor floods the base feed and the pager feed at once? let's say they write one chunk  to all pager indices and additionally they overload the base indices with 3 more extra chunks at places that they want to censor?

SWIPs/swip-191-192.md
```diff
@@ -0,0 +1,266 @@
...
+Applications that need low-latency polling can attach a third feed to the same thread: the new-updates feed.
+
+The new-updates feed is designed to be short-lived.
+
+- each new update on the base thread is also written to the current new-updates feed;
+- the feed changes with epochs, for example one epoch every five minutes;
+- the feed identity is derived from the anythread owner, the topic, and the current epoch boundary;
+- when the epoch changes, a new new-updates feed begins.
+
+Because each such feed exists only for a short interval, it is much less likely to develop large garbage-collected gaps. Even if a small gap appears, looking ahead a few indexes remains cheap. A reader that wants to poll for fresh activity can therefore monitor the current epoch's new-updates feed instead of relying on the first missing index of the long-lived base feed.
```
nugaon, line 253: in general, this time-anchored feed can be used in any situation where the topmost updates are significant and then I still don't see the meaning to use other feed types.

nugaon (commented):
I wanted to comment the missing references in the proposal, but I recognized [my PR about GSOC](https://github.com/nugaon/SWIPs/blob/graffiti-soc/SWIPs/swip-draft_graffiti-soc.md) has disappeared in this repo... not even in the closed PRs. How is that possible?

Anyway, see my comments below.

_edit: I found it by URL: https://github.com/ethersphere/SWIPs/pull/41 , but is it not shown between [pulls](https://github.com/ethersphere/SWIPs/pulls)._

SWIPs/swip-191-192.md
lat-murmeldjur: this limitation exists, but not in a practical sense. to get 1000 chunks in one neighborhood, you only need a batch with depth +10 of the neighborhood depth, it's not prohibitively expensive to get one for even a single actor, and 1000 versions already means that the original ends up behind a hard to digest number of other entries

SWIPs/swip-191-192.md
lat-murmeldjur: it does have an impact on the UX, but only for writers, for loading versions as a reader the time it takes is unaffected, as the reader starts within the narrower range of relevant versions, by calling the get_soc_versions with a set prefix instead of no prefix at the beginning

SWIPs/swip-191-192.md
lat-murmeldjur: first of all thank you for reading through and the valuable feedback, I will try to clarify each point as much as I can

SWIPs/swip-191-192.md
lat-murmeldjur: it's not necessary to use a different batch for one writer writing the base feed and pager feed.

the pager feed only has one purpose. Let's say there was a popular thread 5 years ago, it got entries in the first 1000 base feed indexes, which correspond to the first 32 pager indexes. 5 years later, only one entry survived expiry, it is on index 728 of  the base feed. Now to get that one comment, without the pager feed, one would have to look up the first 728 indexes and then it wouldn't be for sure that further lookahead would not give sporadic results. 

This index is index 23 on the pager feed. Since all but this entries are GC'd on both feeds (as the users used the same batches to write both entries), we can more easily check for sporadic remnants by looking up 32 SOCs (the first 32 pager indexes) to check 1024 base feed indexes. This is more efficient, as it requires 992 less retrieve requests to check the larger range.

SWIPs/swip-191-192.md
lat-murmeldjur: Because a normal user looks up the first free index on the base feed, and decide which pager feed entry to write based on "base feed index / pager size", the entries of a pager feed will fluctuate between 0 and pager size (unless there is an attacker overloading one entry intentionally). This does not mean that a pager feed entry can not have more versions than what the pager size is, it only means normal users will not create more versions for a pager feed soc than pager size, so it will not create a distributed effort by good willing users where any SOC gets an unbounded number of versions. This will limit the "normal usage" versions to pager size.

SWIPs/swip-191-192.md
lat-murmeldjur: We will still efficiently get all versions of all base feed and pager feed indexes based on swip 191. Beyond that, applying the security parameter makes it more expensive for an attacker to generate 100 versions of a new base feed index. If a user notices the soc containing his version is targeted by an attacker, it will be cheaper for that user to write to a new feed index (with zero previous versions) and the attacker will once again need to do magnitudewise more effort to efficiently flood that new index with another 100 versions.

SWIPs/swip-191-192.md
lat-murmeldjur: If a feed has history of 5 months, you would have to check 5 * 31 * 24 * 60 / 5 indexes to get each epoch anchored feed from that period, even if there were only 8 comments in this period. That is an inefficient amount of lookup. Also it's not trivial how much we have to go back in time to get the first such feed with entry. Therefore this construct only helps with live updates. To get historical results it is much more efficient to look up the base and pager feeds.

SWIPs/swip-191-192.md
lat-murmeldjur: This construct is not meant to give more protection against garbage collection for writers. It's sole purpose is to make it very efficient for readers to scan large ranges with sporadic entries. Base and pager feed entry of the same user is expected to GC at the same time, and that is fine. When the base feed entries get sporadic, non-continous, the most efficient way to get the sporadic entries from the higher indexes is to finish base feed lookup until the next pager delimitator one by one, and then look ahead based on pager indexes.

For example, if the pager size is 32, and the first 170 base feed indexes were continous, it's still good to look up each entry until index 192 on the base feed (this is why a  medium pager size makes sense, so we only have to look up another up to 31 entries on the base feed after we find the first gap). From indexes 192, we can continue looking up pager feed index 6 to 38, and those 32 requests will already give us information whether there are any sporadic updates on the next 1024 base feed indexes.

SWIPs/swip-191-192.md
nugaon: I just wanted to correct the phrasing, they need to use the same postage batch (not stamp) so that different updates can expire the same time.

why would I check the 23 index of the pager feed? why the app just don't terminate on the first pager index where it did not find anything?

SWIPs/swip-191-192.md
nugaon: yes, I meant the load time for sending message.

SWIPs/swip-191-192.md
nugaon: if we have this "security parameter" on GSOC, the hostile actor will have the same work to generate messages for that particular address as doing that for two addresses, just with the exponential cost for postage usage. 
in both cases, the original update has the same ambiguity, and updater should do the same work, as well as the hostile actor.

SWIPs/swip-191-192.md
nugaon: yes, so I would use [time-anchored feeds](https://github.com/ethersphere/SWIPs/blob/archive/SWIPs/swip-streaming-of-feeds.md) for latest updates (for not historical results)

PR #92, opened 2026-04-20
SWIP-44: eligibility for sampling and game participation during reserve-expanding pull-sync 

significance:


zelig (approved):
jaw-dropping accuracy, cannot find a problem with this SWIP

SWIPs/SWIP-44.md
```diff
@@ -0,0 +1,321 @@
...
+
+> **Key ordering.** The protocol suspends *syncing* during sampling, rather than
+> blocking *sampling* during syncing. This ensures that eligible nodes always
+> participate in every round.
+
+**Definition 9 (Sampling Window).** The sampling window $[t_s, t_e]$ is the time
+interval during which the node computes its reserve sample for the current
+redistribution round.
+
+**Algorithm 3 (Sync Suspension Protocol).** On entry to the sampling window:
```
acud: while there are still a lot of details for me to digest with this proposal, this part for me is particularly alarming. getting sync to work well is hard enough; why do we want to introduce an architectural change that basically goes against every common sense of having a reasonable UX? do we see ethereum preventing you from submitting a transaction while a block is mined?

i really don't think that continuously turning syncing on and off is a good idea, not at all. also, how do you prevent pushsync chunks from coming in during this time-window? from my perspective all of these considerations must be handled on the persistence abstraction level, not on the protocol level.

significance:
@acud on closer inspection syncing around newly arrived chunks should be out of scope for this change

the intention was to avoid the sync state guarding participation to the game. i have reformulated the swip to amend the guard to be more sympathetic to the requirements and hence obviated the need to pause historical sync in the new bin

the problem with pushsync newly arrived chunks is orthogonal and something to be addressed in a different forum

SWIPs/SWIP-44.md
zelig: yes you are right. i guess it is amended. whether pullsync will still interfere with sampling even in these cases is yet to be checked, cos if it does and the game participation is compromised, then this PR is meaningless

zelig (approved):

acud (approved):

acud:
@significance https://github.com/ethersphere/bee/pull/5568 would be good if you could review this

PR #19, opened 2019-08-23
SPEC: BZZ Handshake

nolash:
Adds a high level description of the handshake protocol; which function it has, what is included in it and how nodes should react to it.

As a curiosity feel free to inspect the bzz handshake _hack_ for `nim-eth`, that successfully negotiates towards a node running the reference golang implementation (and appears in its kademlia). https://github.com/oskarth/nim-eth/pull/2

---

As we are using the SWIP review process for these submissions, the documents are formatted to the SWIP template. It is, however, not an _improvement proposal_ and thus only the **Specification** section is relevant.

Also, please consider these submissions in the general context of a full specification document. For this reason, internal document reference links in the markdown douments may very well not point to anywhere.

If the full context is needed, please see the PDF build of current [node implementer spec master branch](https://github.com/ethersphere/node-implementer-spec) state, which can be found on [this swarm-feed address](https://swarm-gateways.net/bzz:/dcad67ce1b19aa80d8bb2c1e5c2244c2e67f3f0f894cf863ed5522d0943e4c14/)

PR #94, opened 2026-05-09
SWIP-XXXX: Swarm Provider API (window.swarm)

flotob:
## Summary

This PR proposes a new draft SWIP defining a standard JavaScript provider API — `window.swarm` — that lets web pages request access to a user's Swarm (Bee) node for publishing data, uploading files, managing mutable feeds, reading and writing indexed feed entries, and introspecting their own feed records. It follows the request/response pattern established by [EIP-1193](https://eips.ethereum.org/EIPS/eip-1193) for Ethereum providers, adapted for Swarm's publishing and feed primitives.

The draft lives at [`SWIPs/swip-draft_provider_api.md`](SWIPs/swip-draft_provider_api.md) and will be renamed to `swip-XXXX.md` once a number is assigned.

## Motivation

Today, publishing to Swarm from a web app requires either:

1. Direct HTTP calls to a locally running Bee node — exposing the **full** API surface (including administrative, debug and staking endpoints) to any page that can reach `localhost`, or
2. Server-side publishing infrastructure — defeating the point of decentralized hosting.

Neither is acceptable for a user-sovereign web. Users need a way to grant web apps controlled, permissioned access to Swarm publishing — the way `window.ethereum` lets pages interact with a wallet without exposing private keys. Without a standard, every Swarm-enabled browser or extension will invent its own API, fragmenting the ecosystem.

## What the SWIP specifies

- **Provider object** (`window.swarm`) injected before `DOMContentLoaded`, with `request()`, convenience wrappers, and an event interface.
- **Ten RPC methods** covering: connection (`swarm_requestAccess`), capability discovery (`swarm_getCapabilities`), publishing (`swarm_publishData`, `swarm_publishFiles`, `swarm_getUploadStatus`), mutable feeds (`swarm_createFeed`, `swarm_updateFeed`), indexed feed entries (`swarm_writeFeedEntry`, `swarm_readFeedEntry`), and origin-scoped feed introspection (`swarm_listFeeds`).
- **Permission model** with explicit user consent, **origin normalization** (so all dweb content served from `127.0.0.1` doesn't collapse into one permission scope), and **app-scoped feed identities** (each origin's feeds signed by a per-origin derived key, so dApps can't impersonate the user's main wallet).
- **Read-only methods on public Swarm data** (feed entry reads, listing the calling origin's own feeds) require **no permission grant**, since the same data is fetchable via any Bee gateway and gating it would be friction without security benefit.
- **Exact-match requirements** on indexed feed reads/writes (with implementation notes on Bee's `GET /feeds/{owner}/{topic}` at-or-before semantics and the SOC-address fallback) so overwrite protection and explicit-index reads are safe.
- **Capability-advertised limits** (`maxDataBytes`, `maxPathBytes`) so future implementations supporting PAX tar can advertise larger paths without breaking the spec.
- **Security considerations** covering origin trust, resource exhaustion, iframe behavior, key material isolation, and temp artifact cleanup.
- **Future extensions** explicitly out of scope for 1.0 (`capabilitiesChanged` event, `preferredIdentityMode`, encoding parameter for reads, promotion path for `specVersion`).

## Reference implementations

This is not a paper spec — both sides of the API are already implemented and exercise the full surface:

- **Provider:** [Freedom Browser](https://github.com/solardev-xyz/freedom-browser) ([PR #19](https://github.com/solardev-xyz/freedom-browser/pull/19)) injects `window.swarm` into webview contexts, with main-process IPC enforcement, origin normalization, app-scoped feed keys, and per-origin permissions.
- **Consumer dApp:** [Swarmit](https://github.com/flotob/swarmit) — a decentralized message board — uses `swarm_requestAccess`, `swarm_publishData`, `swarm_createFeed`, `swarm_writeFeedEntry`, `swarm_readFeedEntry`, and `swarm_listFeeds` for its full publishing and profile-discovery pipeline.

The implementation work surfaced several spec issues that have already been folded back into this draft (USTAR tar's 100-byte path limit, Bee's at-or-before feed lookup semantics, the SOC-address endianness trap, permission gating for public reads).

## Status

- **Status:** Draft — opening this PR to register the SWIP and begin discussion.
- **Developer Documentation:** [Swarm Provider JavaScript API](https://swarm-api.freedombrowser.eth.limo/)
- **Discussion:** [Swarm Discord — SWIPs channel](https://discord.com/channels/799027393297514537/1239813439136993280).
- Happy to iterate on naming, scope, and any specifics before this is endorsed. Filename will be renamed from `swip-draft_provider_api.md` to `swip-XXXX.md` once a number is assigned.

### DIFF
--- a/SWIPs/SWIP-draft_swarm_provider_api.md
+++ b/SWIPs/SWIP-draft_swarm_provider_api.md
@@ -1,0 +1,545 @@
+---
+SWIP: <to be assigned>
+title: Swarm Provider API (`window.swarm`)
+author: Florian Glatz (@heckerhut)
+discussions-to: https://discord.com/channels/799027393297514537/1239813439136993280
+status: Draft
+type: Standards Track
+category: Interface
+created: 2026-04-03
+---
+
+## Simple Summary
+
+A standard JavaScript API (`window.swarm`) that enables web pages to request access to a user's Swarm node for publishing data, uploading files, and managing mutable feeds — with user consent and origin-scoped permissions.
+
+## Abstract
+
+This SWIP defines a browser-injected JavaScript provider object (`window.swarm`) that allows web applications to interact with a user's local Swarm (Bee) node. The API follows the request/response pattern established by [EIP-1193](https://eips.ethereum.org/EIPS/eip-1193) for Ethereum providers, adapted for Swarm's publishing and feed primitives. It specifies seven RPC methods covering connection, capability discovery, data/file publishing, upload tracking, and mutable feed management. The provider includes a permission model with explicit user consent, origin isolation, and upload size limits.
+
+## Motivation
+
+Today, publishing to Swarm from a web application requires either:
+
+1. Direct HTTP calls to a locally running Bee node API (exposing the full API surface, including administrative endpoints, to any web page that can reach `localhost`), or
+2. Server-side publishing infrastructure (defeating the purpose of decentralized hosting).
+
+Neither approach is suitable for a user-sovereign web. Users need a way to grant web applications controlled, permissioned access to Swarm publishing capabilities — similar to how `window.ethereum` ([EIP-1193](https://eips.ethereum.org/EIPS/eip-1193)) lets web applications interact with a user's Ethereum wallet without exposing private keys.
+
+Without a standard, each Swarm-enabled browser or extension will invent its own API. This fragments the ecosystem: dApp developers must write implementation-specific code, and users can't switch clients without breaking functionality. A standard `window.swarm` provider ensures interoperability across implementations.
+
+### Design Goals
+
+- **User consent first.** No operation executes without explicit permission.
+- **Minimal surface.** Only publishing and feeds are exposed — not administrative, debug, or staking endpoints.
+- **Origin isolation.** Each web origin gets isolated permissions and feed identities.
+- **Implementation freedom.** The spec defines the interface, not how the host manages nodes, keys, or storage internally.
+
+## Specification
+
+### Provider Object
+
+Conforming implementations MUST inject a `window.swarm` object into the JavaScript context of web pages. The object MUST be available before the page's `DOMContentLoaded` event fires.
+
+#### Detection
+
+```javascript
+if (typeof window.swarm !== 'undefined') {
+  // Swarm provider is available
+}
+```
+
+#### Properties
+
+| Property | Type | Description |
+|---|---|---|
+| `isFreedomBrowser` | `boolean` | **(OPTIONAL, implementation-specific)** Implementations MAY include a boolean property identifying themselves. This is NOT part of the standard interface and MUST NOT be relied upon for feature detection. Use `swarm_getCapabilities` instead. |
+
+> **Note:** Future versions of this specification may define a standard `isSwarmProvider` property or a capability-based detection mechanism.
+
+### Request Interface
+
+The provider MUST expose a `request` method:
+
+```javascript
+window.swarm.request({ method: string, params?: object }): Promise<any>
+```
+
+- `method` — (Required) A string identifying the RPC method to invoke.
+- `params` — (Optional) A plain object containing method-specific parameters.
+
+The method MUST return a `Promise` that resolves with the method's result, or rejects with an error object.
+
+### Convenience Methods
+
+Implementations SHOULD expose convenience wrappers for each standard method:
+
+```javascript
+window.swarm.requestAccess()                     // swarm_requestAccess
+window.swarm.getCapabilities()                    // swarm_getCapabilities
+window.swarm.publishData({ data, contentType })   // swarm_publishData
+window.swarm.publishFiles({ files })              // swarm_publishFiles
+window.swarm.getUploadStatus({ tagUid })          // swarm_getUploadStatus
+window.swarm.createFeed({ name })                 // swarm_createFeed
+window.swarm.updateFeed({ feedId, reference })    // swarm_updateFeed
+```
+
+These MUST be equivalent to calling `window.swarm.request()` with the corresponding method name.
+
+### Events
+
+The provider MUST support event subscription:
+
+```javascript
+window.swarm.on(event: string, handler: Function): this
+window.swarm.removeListener(event: string, handler: Function): this
+window.swarm.removeAllListeners(event?: string): this
+```
+
+#### Standard Events
+
+| Event | Data | Description |
+|---|---|---|
+| `connect` | `{ origin: string }` | Emitted when the user grants access. |
+| `disconnect` | `{ origin: string }` | Emitted when the user revokes access or the provider disconnects. |
+
+### Error Format
+
+Errors MUST follow the [JSON-RPC 2.0 error format](https://www.jsonrpc.org/specification#error_object):
+
+```javascript
+{
+  code: number,
+  message: string,
+  data?: any
+}
+```
+
+#### Standard Error Codes
+
+| Code | Name | Description |
+|---|---|---|
+| 4001 | User Rejected | The user denied the request. |
+| 4100 | Unauthorized | The origin has not been granted access. Call `swarm_requestAccess` first. |
+| 4200 | Unsupported Method | The requested method is not recognized. |
+| 4900 | Node Unavailable | The Swarm node is not running, not ready, or lacks usable postage stamps. |
+| -32602 | Invalid Params | Missing or invalid method parameters. |
+| -32603 | Internal Error | An unexpected error occurred in the provider. |
+
+Error codes 4001, 4100, 4200, and 4900 are aligned with the [EIP-1193](https://eips.ethereum.org/EIPS/eip-1193) and [EIP-1474](https://eips.ethereum.org/EIPS/eip-1474) error code ranges.
+
+---
+
+### Methods
+
+#### `swarm_requestAccess`
+
+Requests the user's permission to interact with their Swarm node from this origin. Implementations MUST present a user-visible consent prompt. The user MUST be able to deny access.
+
+**Params:** None.
+
+**Result:**
+
+```javascript
+{
+  connected: true,
+  origin: string,          // The normalized origin that was granted access
+  capabilities: string[]   // e.g. ["publish"]
+}
+```
+
+**Errors:** `4001` if the user rejects.
+
+**Behavior:**
+- Calling `swarm_requestAccess` when already connected SHOULD return the existing connection state without re-prompting.
+- After a successful `swarm_requestAccess`, the provider MUST emit a `connect` event.
+
+---
+
+#### `swarm_getCapabilities`
+
+Returns the current capabilities of the provider for this origin. Does NOT require prior access — can be called before `swarm_requestAccess` for feature detection.
+
+**Params:** None.
+
+**Result:**
+
+```javascript
+{
+  canPublish: boolean,     // true if connected AND node is ready
+  reason: string | null,   // null if canPublish is true, otherwise a reason code
+  limits: {
+    maxDataBytes: number,  // Maximum payload size for swarm_publishData
+    maxFilesBytes: number, // Maximum total size for swarm_publishFiles
+    maxFileCount: number   // Maximum number of files per swarm_publishFiles call
+  }
+}
+```
+
+**Reason codes:**
+
+| Code | Meaning |
+|---|---|
+| `"not-connected"` | Origin has not called `swarm_requestAccess`. |
+| `"node-stopped"` | Bee node is not running. |
+| `"ultra-light-mode"` | Node is in ultra-light mode (browse only, cannot publish). |
+| `"node-not-ready"` | Node is running but not yet synced/ready. |
+| `"no-usable-stamps"` | No postage stamps with remaining capacity. |
+
+---
+
+#### `swarm_publishData`
+
+Upload a single blob of data to Swarm.
+
+**Params:**
+
+| Field | Type | Required | Description |
+|---|---|---|---|
+| `data` | `string \| Uint8Array \| ArrayBuffer` | Yes | The content to upload. |
+| `contentType` | `string` | Yes | MIME type (e.g. `"text/html"`, `"application/json"`). |
+| `name` | `string` | No | Display name for the upload. |
+
+**Result:**
+
+```javascript
+{
+  reference: string,   // 64-character hex Swarm reference
+  bzzUrl: string       // "bzz://<reference>"
+}
+```
+
+**Errors:** `4100` if not connected, `4900` if node unavailable, `-32602` if params invalid or payload exceeds `maxDataBytes`.
+
+**Behavior:**
+- Implementations SHOULD prompt the user for per-publish consent unless the user has opted into auto-approve for this origin.
+- The implementation selects an appropriate postage batch automatically — dApps do not manage batches.
+- Uploads MUST be pinned on the local node.
+
+---
+
+#### `swarm_publishFiles`
+
+Upload a collection of files as a Swarm manifest (directory).
+
+**Params:**
+
+| Field | Type | Required | Description |
+|---|---|---|---|
+| `files` | `Array<FileEntry>` | Yes | Array of file objects. |
+| `indexDocument` | `string` | No | Path of the default document (e.g. `"index.html"`). Must match an entry in `files`. |
+
+**FileEntry:**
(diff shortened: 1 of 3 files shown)


agazso:
It would be good if chunk level operations were supported in the API, because with those every higher level Swarm APIs can be implemented (feeds, manifest, ACT etc.), even ones that are not supported at the moment by Bee (e.g. epoch feeds).

The API could be `window.swarm.publishChunk({ chunkPayload, span, options }) ` where `chunkPayload` is the chunk payload with type `Uint8Array | string`, and an optional `span`  that can be omitted and it defaults to the payload length (in bytes). `options` would be also optional, and it can contain other fields controlling advanced functionality (pin, encryption, tag, deferred, redundancy). 

Similarly there would be a `window.swarm.readChunk({ reference, options })` which would return the chunk data. The `reference` could be either `Uint8Array | string` and it could contain a regular or an encrypted reference. The `options` would be optional and it can contain other fields controlling advanced functionality (redundancy, ACT).

There are also Single-Owner Chunks. Currently Swarm does not have a way to distinguish different chunk types, and it does it with heuristics, but there are documented issues and discrepancies about it.
See: 
- https://github.com/ethersphere/SWIPs/pull/67
and
- https://github.com/ethersphere/bee/pull/5445

If this could be fixed at the Swarm level, then `publishChunk` would be enough for publishing and only a `readSingleOwnerChunk` would be needed additionally, otherwise also a `publishSingleOwnerChunk` would be necessary. But maybe that's out of the scope of this SWIP already.

flotob:
Thanks @agazso — this ended up being very useful feedback. The latest push (`d42af03`) takes the chunk tier substantively on board.

**Chunk methods added:** `swarm_publishChunk` / `swarm_readChunk` for CACs, and `swarm_writeSingleOwnerChunk` / `swarm_readSingleOwnerChunk` for SOCs. The chunk type is declared by the method name rather than inferred, which avoids depending on the SOC/CAC heuristic ambiguity discussed in #67 and ethersphere/bee#5445.

**Type validation on reads:** providers must verify returned bytes against the requested chunk type: BMT/address validation for CACs, and SOC signature recovery plus `keccak256(identifier || owner)` derivation for SOCs. A mismatch returns `chunk_type_mismatch`.

**Signing identity discovery:** added `swarm_getSigningIdentity`, so dApps can pre-compute SOC addresses, e.g. for epoch-feed construction, before committing to a specific write. It also gives a clean bootstrap path for the feed/signing permission grant.

**`options` exists, but is intentionally narrow in v1:** chunk methods accept an `options` object, but no fields are standardized yet. Unknown fields must return `unsupported_option`. ACT, encryption, redundancy, progress tags, cache behavior, and deferred upload are listed as future-extension candidates pending a precise mapping to Bee’s endpoint semantics.

**SOC mutability wording corrected:** same `(owner, identifier)` rewrites are undefined/discouraged; mutable structures should use distinct identifiers, with `swarm_writeFeedEntry` retaining overwrite-rejection semantics for the journal/feed use case.

A few related changes fell out of this: permission-free reads now have mandatory per-origin rate/bandwidth limits with a `rate_limited` reason; `span` is now `number | bigint` for u64-safe custom BMT spans; and `4001` vs `4100` is now consistent across publish/feed permission prompts.

I also looked at SwarmID while shaping this. My current read is that `window.swarm` should provide the node-access, permission, origin, postage-selection, and signing boundary, while libraries such as SwarmID can build identity/encryption/ACT/higher-level structures on top of the chunk substrate. Happy to coordinate on what a SwarmID adapter would need from this interface.

Reference implementations for the new chunk tier and `swarm_getSigningIdentity` will follow in a subsequent Freedom Browser PR; the existing high-level methods are already implemented there. Feedback on the new surface before that work starts is very welcome.

flotob:
Hi @agazso, quick follow-up: we implemented the chunk-level provider API discussed here in Freedom Browser.

Implementation PR: https://github.com/solardev-xyz/freedom-browser/pull/81

It includes:
- CAC methods: `swarm_publishChunk` / `swarm_readChunk`
- SOC methods: `swarm_writeSingleOwnerChunk` / `swarm_readSingleOwnerChunk`
- `swarm_getSigningIdentity`
- permission-free read budgets for public reads
- explicit CAC/SOC type validation with `chunk_type_mismatch`
- structured provider error reasons

We also smoke-tested the CAC and SOC paths against Bee. The smoke test caught an SOC upload endpoint issue, which is now fixed: SOC writes use Bee's `/soc/{owner}/{identifier}` API and reads work both by SOC address and by owner+identifier.

Thanks for the idea that led to this amendment; it translated cleanly into the provider implementation.

flotob:
Quick additional follow-up: we also built a first experimental consumer library on top of the chunk-level API:

https://github.com/solardev-xyz/swarm-kit

The interesting part is that this validates the original reason for adding raw CAC/SOC access: higher-level Swarm data structures can now be implemented outside the provider/Bee API surface.

`swarm-kit` currently includes helpers/primitives for:

- CAC/SOC byte, text, and JSON helpers
- deterministic SOC identifier derivation
- object/chunk-graph storage above one raw chunk
- epoch feeds
- indexed SOC streams
- hash chains
- multi-writer feeds
- DID-style / revisioned owner records
- encrypted object helpers
- signed document records
- commit/reveal pairs
- provider compliance checks for `window.swarm`

The playground and compliance harness have been smoke-tested in Freedom Browser against the implemented `window.swarm` chunk methods.

This is still experimental, but it is a useful data point for the SWIP: the chunk tier does seem to move many future Swarm patterns from “new provider/Bee method needed” into “library-level convention.”

flotob:
I've pushed a companion draft to this branch — **`swip-draft_provider_api_messaging.md`** — proposing a **messaging extension** (PSS + GSOC) to the provider API, and wanted to surface it here for feedback since it builds directly on this spec.

**Why an extension:** the v1 surface is entirely request/response — there's no way for a dApp to *receive* asynchronous data (the only push signals are `connect`/`disconnect`). That's fine for publishing and mutable feeds, but it structurally can't support real-time applications (collaborative editing, chat, presence), which need a channel over which the node delivers messages as they arrive. Polling feeds gets you async sync but never interactive latency, and it can't express PSS's push-only encrypted point-to-point delivery at all.

**What it adds** (four methods + one event):

- `swarm_subscribe` / `swarm_unsubscribe` — the first streaming surface, using the EIP-1193 `eth_subscribe` pattern (subscription id + a `message` event) to stay idiomatic.
- `swarm_sendPss` — encrypted point-to-point to a recipient's PSS public key.
- `swarm_sendGsoc` — topic broadcast; any subscriber to the derived GSOC address receives it.
- `swarm_getMessagingIdentity` — discloses the origin's PSS public key + overlay so peers can address it.

It reuses the core spec's consent model (a new **messaging permission tier**), origin-scoping, base64 payload encoding, and error format, and is purely additive — feature-detected via `getCapabilities().features.includes("messaging")`.

**A few design points I'd especially value feedback on:**

1. **The streaming shape.** I went with EIP-1193's subscription-id + `message` event rather than per-method callbacks, to keep a single streaming idiom. Reasonable, or is there a preferred convention?
2. **GSOC send as a distinct method** rather than a mode of `swarm_writeSingleOwnerChunk` — because GSOC needs a *mined* owner key for neighborhood placement, not the origin's fixed signing identity. I think a separate method keeps the core SOC-write contract clean, but I'm open to folding it in.
3. **Delivery semantics.** I made "best-effort, unordered, at-least-once" normative and prominent, steering applications toward CRDT/idempotent designs and toward the durable feed layer for authoritative state (feeds = source of truth, messaging = live propagation).
4. **GSOC overwrite behavior for delta streams** — I flagged that where every update matters, senders should vary the address per message or use PSS, since a GSOC address holds a single latest value. Want to make sure that framing matches implementers' expectations.

This isn't speculative — it's designed against a working light-node PSS/GSOC implementation (send **and** receive on a light node, interop-tested against bee and verified on mainnet), and an informative "Reference Endpoints" section maps each method onto the gateway endpoints that back it. Happy to split it into its own PR if that's cleaner for review; I kept it here since it directly extends this spec.

flotob:
A self-review pass on the messaging draft turned up a few things worth fixing before anyone else hit them — pushed as 49eab20:

1. **GSOC topic→address interop, stated honestly.** The draft required deterministic derivation and gestured at bee-js `gsocMine` for cross-implementation interop — but determinism alone doesn't pin the address. Two conforming providers must also agree on the identifier derivation, the target-neighborhood derivation, the mining search order, *and* the required proximity; differing on any one yields a different address for the same topic. The spec now guarantees convergence **within** an implementation, says cross-provider apps exchange the resolved `address` (returned by both `swarm_subscribe` and `swarm_sendGsoc`) and use the raw `address` param, and defers a normative derivation profile to Future Work. This is also the design question I'd most value input on: **should v1 pin a normative topic→address profile** (so `"room:doc-42"` means the same address everywhere), or is the address-exchange escape hatch acceptable until implementations converge on one?

2. **`swarm_getMessagingIdentity` no longer returns the full node overlay.** The overlay is node-global — identical for every origin — so returning it would hand any two messaging-granted dApps a cross-origin correlation handle that also pinpoints the user's node's network position. PSS routing only needs a prefix, so the method now returns a truncated `pssTarget` (≤ `maxTargetDepth` bytes); `swarm_sendPss`'s `targets` is now required and documented as the recipient's `pssTarget`. New Security Considerations entry covers the correlation reasoning.

3. **Smaller normative fixes:** added the missing `invalid_target` row to the error-reason table (it was referenced but never defined); reworded mid-stream failure signalling, which previously suggested emitting an error inside the `message` event whose schema has no error variant (implementations MUST document degraded-pipeline behavior; normative signalling stays in Future Work); noted that byte-identical sends on the same key are indistinguishable at the transport layer and may coalesce (apps needing distinguishable repeats include a nonce/sequence); `features`/limits are MUST-available pre-grant, matching the core spec's permission-free `getCapabilities`; and a SHOULD-level per-origin send rate cap under auto-approve, since sends consume postage.

None of this changes the four methods or the streaming shape — it tightens claims to what the mechanics actually guarantee.

flotob:
Small follow-up to the messaging draft (19cfb9d): hardening the reference light-node implementation surfaced two spec gaps worth closing before they bite an implementer.

1. **Payload bounds were only half-specified.** The draft bounded `data` from above (`payload_too_large`) but said nothing about empty payloads — and the two transports genuinely differ there: a GSOC update is a single-owner chunk, and bee's chunk layer **rejects an empty SOC payload**, so an empty broadcast can't be expressed and the provider must not sign a chunk the network refuses (`swarm_sendGsoc` is now `1..maxMessageBytes` with a new `invalid_payload` reason). PSS, by contrast, carries an explicit length in the trojan framing, so a zero-byte message is well-formed — `swarm_sendPss` now explicitly allows it (handy for pings/presence signals). Test cases added for both directions.

2. **Node capacity ≠ per-origin cap.** `too_many_subscriptions` covers the origin exceeding its own `maxSubscriptions` — the dApp caused it and can fix it. But a receive pipeline is a *node-wide* resource: an implementation bounds how many distinct neighborhoods it will pull concurrently across **all** origins (our reference implementation allows 8), and that pool can be exhausted by *other* origins. The draft now requires that case to surface as a **retryable `4900`** naming subscription capacity — not as a parameter error — so dApps back off or degrade to feed polling instead of treating it as a permanent refusal. The limits table notes why the aggregate bound isn't advertised per-origin (it's shared state, not a promise any one origin can hold).

Both came out of adversarial review of the implementation rather than armchair speculation, which is exactly the feedback loop this draft is meant to have with running code.

flotob:
Two related amendments in `abacf6b`, both from **Viktor Trón's** review of the light-node receiver behind this spec — and both surfacing something the draft under-specified.

**1. PSS mining depth is now a real constraint (`pssTarget` / `sendPss.targets` → 2..`maxTargetDepth` bytes, not 1..).** The prefix length is the number of leading bits a trojan is mined to match, and it's bounded on *both* sides:
- **Floor (storability):** a chunk mined shallower than the network's storage depth lands outside every neighborhood reserve — so it's *not stored*, and the send "succeeds" (accepted for upload) but is undeliverable. A 1-byte (L=8) target is below the public network's depth. The draft's "1–`maxTargetDepth` bytes, your choice" was a trap at the shallow end; 1-byte targets are now rejected (`invalid_target`).
- **Ceiling (mining cost):** ~2^L hashes, capped by `maxTargetDepth`.

A new normative **"PSS mining depth"** section explains both, plus the receiver-cost geometry Viktor pointed out: each extra prefix bit *halves* the candidate traffic a light-node receiver must sift.

**2. A mailbox — `swarm_subscribe({ history: true })` (PSS only).** On subscribe, the provider delivers messages sent *before* the subscription — offline delivery — bounded by storer retention and an implementation-defined lookback. Delivery-semantics section spells out the bounded/best-effort guarantees; `receivedAt` is recovery time, not send time; GSOC has no history (a SOC has a latest value, not a backlog). New test case shows the before/after.

**Why they're in one amendment:** the two compose. At the current 2-byte prefix a mailbox can only recover a *recent* window (the trojan bins are large). At a 3-byte (L=24) prefix the whole neighborhood trojan backlog is ~1–2K chunks — small enough to sweep completely on subscribe — so history recovers *all* offline messages. Deeper mining makes the mailbox complete. Both are **implemented and mainnet-verified** on the reference light node: the bin-selection correction, and the mailbox (4/4 messages sent with nobody subscribed, then recovered by a `history` subscribe).

**Open question for the community — worth a decision:** the draft still doesn't pin a single normative `L`. 2 bytes maximizes compatibility with today's senders; 3 bytes is materially better for receiver cost *and* unlocks the complete mailbox. This feels like a network-wide parameter that belongs alongside the storage-depth conventions in the core protocol docs, not only in this provider spec — flagged as such. Guidance from those closest to the protocol (Viktor?) on where to standardize it, and on 2 vs 3, would settle it.

flotob:
**Resolved: `L = 24` (3-byte targets)** — Viktor Trón ruled on the open question from the last revision. Committed as `de1bddf`.

The spec now adopts 24 bits normatively: senders SHOULD mine 3-byte targets, receivers assume `L = 24`, 2 bytes stays the hard storability floor but is deprecated for interop. The delivery-semantics section's complete-mailbox claim is pinned to L=24 (the whole neighborhood backlog is ~1–2K chunks there — small enough to sweep entirely on subscribe).

Both directions are now **implemented and mainnet-verified** on the reference light node at L=24: live PSS 4/4, and the mailbox 4/4 (messages sent with nobody subscribed, recovered by `?history=true`).

I also added an **informative mining-cost note**, because it surfaced in testing and matters for implementers: 24-bit mining is ~256× a 2-byte mine. Server-side that's a few seconds; on a constrained mobile sender it's material, and it occasionally trips a 60s send timeout even in an optimized build (geometric variance in the search). The spec now recommends mining off the request's critical path rather than blocking a send on it. This is the one real cost of the deeper convention — it buys ~256× cheaper *reception* and complete offline delivery at the price of more *sender* work — and it's a UX concern, not a protocol one.

Left a note that `L`, governing storage and retrieval for every PSS participant, is arguably a core-protocol parameter whose broader normative home is the Swarm protocol docs; this SWIP fixes it at the provider boundary.

flotob:
**Reverted the convention to `L = 16` (2-byte default)** — `765f8e6`. The previous "adopt L=24" revision is superseded after further discussion.

The correction that drove it: a deeper prefix only reduces reception cost / enlarges the recoverable mailbox for a receiver **deeply resident** in the target neighborhood (covering peers at proximity ≥ L). A light node — which is what this extension targets — pulls the same bins and downloads the same candidates regardless of L (measured identical at 16 and 24 on a reference light node). So for light-node deployments, deeper mining adds ~256×/byte of *sender* work with no *receiver* benefit, and the `history` mailbox is a recent-history window at any L. Combined with the fact that PSS already carries an economic anti-spam gate (a postage stamp per send), the extra proof-of-work isn't worth taxing every mobile sender for.

2 bytes remains the hard storability floor; deeper mining is opt-in where sender and receiver privately agree. The reception-cost and delivery-semantics sections are corrected (no more "complete backlog at L=24" claim without the deep-residency caveat).

I kept the **network-level question open and explicit**: whether Swarm *should* mandate a deeper prefix as a deliberate anti-spam / anti-free-riding PoW on senders — distinct from postage — is a protocol-incentive decision raised by the protocol designer and still under discussion. Its normative home is arguably the core protocol docs, not this provider extension; the SWIP fixes the provider default at 16 and will track any core-level ruling.

Issue #86, opened 2026-02-27
Rapid storage price changes adversely impact UX/DX


awmacpherson:
## Strategic context

Pricing UX problem impacting cost predictability and hence viability for professional use cases.

## Problem statement

Swarm's dynamic storage pricing means that users cannot reliably forecast their costs per byte over an interval greater than a single round (twelve minutes and forty seconds). Equivalently, one cannot specify a fixed TTL for a storage quota at purchase time.

If one's goal is, say, to host some dataset for 90 days, the most cost-effective way to do this on Swarm today is to purchase a short-dated quota and keep it topped up as long as one needs, paying whatever the cost is at the time. One can estimate what the cost of a 90-day quota would be and pay that amount up-front, but that has the risk of overshooting (hence wasting money) or undershooting (hence needing a top-up anyway, or expiring early).

Concretely, the current state of price updates presents two challenges:

* The price updates are much more frequent than the intervals over which people usually want to make infrastructure spending decisions.
* After compounding, price changes over those intervals are much larger than what developers are used to.

## Suggested approach

Before we can even evaluate candidate solutions to this issue, some basic questions need to be answered:

* *Does the current system work?* Assess the extent to which the system actually is "dynamically self-regulating" [BoS 3.4.5]. Evaluate the strength of the effect of price adjustments in practice on supply and demand over different timescales in Swarm. Make comments about system stability.
* *What are the costs of the current approach?* Assess potential externalities of the dynamic price adjustment mechanism. Try to quantify its effect on users.
* *What properties should storage pricing have? How will we know if we've solved this problem?* Formulate objectives for pricing range and stability expectations based on market research, target audience, and user feedback.
* *How much and what type of intervention are we prepared to consider?* The current mechanism cannot know about the current goals of Swarm and has extremely limited capacity to respond to the prevailing market context. Sometimes, new research or shifting goals or market conditions may reveal that a different design or parameter tuning is preferable to the deployed version. To realise this improvement, intervention is needed. How often are we prepared to intervene, and what types of interventions are allowed? Even if the long-term goal is for systems to pass the "walk-away test," i.e. be zero-intervention, can we accept a "high-intervention" solution as a stop-gap solution to UX problems while we work towards a more sustainable approach?

kosmostasis:
[...]

PR #95, opened 2026-05-14
Queued updates and withdrawable stake

awmacpherson:
Merged spec from SWIPs 40 and 41, including everything except revert of SWIP-20. Updated to align with implementation at https://github.com/ethersphere/storage-incentives/pull/309.

### DIFF
--- a/SWIPs/swip-40+41.md
+++ b/SWIPs/swip-40+41.md
@@ -1,0 +1,938 @@
+---
+SWIP: unassigned
+title: Queued updates and withdrawable stake
+author: Andrew Macpherson (@awmacpherson)
+discussions-to: https://discord.gg/Q6BvSkCv (Swarm Discord)
+status: WIP
+category: Core
+created: 2025-10-10
+assumes: SWIP-xx (Return to single-parameter stake record)
+---
+
+# Queued updates and withdrawable stake
+
+## Abstract
+
+Introduce per-owner parallel FIFO queues that add parametrised delays to all updates to stake balances and metadata (i.e. height and overlay address). This replaces the 2-round thaw imposed on stakers after any change to stake balance or metadata. Make stake fully withdrawable except while frozen, subject to a queued delay.
+
+## Motivation
+
+### Withdrawable stake
+
+* Swarm's staking model does not generally allow withdrawal of stake, except during smart contract migrations. 
+* There is an intricate notion of "excess stake" that can be withdrawn, after [SWIP-20](https://github.com/ethersphere/SWIPs/blob/master/SWIPs/swip-20.md) "improved staking." This concept substantially complicates the codebase — 1/3 of the tests for the stake registry are testing this function — 
+* The fact that Swarm stakers generally cannot recover their principal except via revenue makes staking a risky prospect, akin to a small venture investment. It's likely that only larger operators will be prepared to take such risks, exacerbating inequality among node operators. Conversely, the current capacity of the investment is likely too small to attract the interest of sophisticated operators.
+* It also goes against the common understanding of what "stake" is.
+* Making stake recoverable actually gives the system more leverage over node operators to behave well, potentially enhancing system service quality — the system can threaten penalties over the principal, instead of only over future revenue.
+
+An option to fully withdraw stake under typical network conditions makes staking a much more attractive, low-risk investment opportunity. Because principal is not at risk, it improves the accessibility of staking to risk-averse operators such as those with worse access to capital. It also makes the opportunity easier to compare on a like-for-like basis with other staking systems. For a more detailed analysis, see [here](https://mirror.xyz/shtuka.eth/qQnVGyNL7viiS5iLizSVL_0eTTMYGavl3Kb77XiaBxk).
+
+### Filtering short-termist operator behaviour
+
+Operators may be incentivised to make changes to their commitments based on short term signals such as market volatility, leading to excessive churn of stake positions. Imposing a wait period $X$ makes it impossible to execute a strategy based on a holding period of less than $X$.
+
+### Advance signalling of service changes
+
+Some operations — for example, height reduction or exit — result in a node's exit from or reduced financial commitment to a neighbourhood, negatively affecting the storage service. In an ideal world, any reduction of service commitment from one node would be compensated by another node arriving to take its place. However, since there is currently no reliable way to predict changes in service commitment, this hand-over cannot happen without at least a short service disruption. Introducing a mandatory "unwinding" delay before such changes come into effect, during which nodes continue to participate under their previous commitments, would provide a reliable signal allowing for smoother handoff of responsibilities between outgoing and incoming nodes.
+
+A mandatory wait period to exit a stake position would be familiar from many types of risky investment, where delays are a standard measure to facilitate orderly unwinding of positions.
+
+### Stake record update induced freeze
+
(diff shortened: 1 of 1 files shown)


significance:
thank you for the SWIP 🙇 

as discussed, please can you emphasise the rationale behind withdrawal delay i.e. encouraging predicability / reducing unpredictability

in this vein i think we have agreement that limiting the amount of times an overlay can be changed over some similar period would also be beneficial in this regard without incurring too much friction in terms of neighbourhood rebalancing

i would suggest it were set at once every half or quarter the period of the withdrawal delay, but certainly less, to allow for some margin of error in coordination of node repositioning 

it was also mooted that 14 days could well be sufficient delay and that the exact value of that constant and/or mode and means of tuning it would benefit from some more discussion

awmacpherson:
> as discussed, please can you emphasise the rationale behind withdrawal delay i.e. encouraging predicability / reducing unpredictability

Sure, I will expand on this.

> in this vein i think we have agreement that limiting the amount of times an overlay can be changed over some similar period would also be beneficial in this regard without incurring too much friction in terms of neighbourhood rebalancing
> 
> i would suggest it were set at once every half or quarter the period of the withdrawal delay, but certainly less, to allow for some margin of error in coordination of node repositioning

While I think this is a reasonable direction to explore due to the sync costs excessive node churn could impose on the network, in terms of both motivation and implementation is it independent of the present SWIP. This SWIP does not make any change to the rate at which one can change neighbourhoods (whether staked or not). So I'd argue that this research direction should not block merging this PR. Would you agree?

> it was also mooted that 14 days could well be sufficient delay and that the exact value of that constant and/or mode and means of tuning it would benefit from some more discussion

Regardless of what value we choose for the delay parameter now, we ought to establish processes to monitor the behaviour of the network and determine whether the delay parameter is performing as desired or if it ought to be changed. I'll work on adding something on that to the SWIP too.

If we are concerned about committing to a particular withdrawal notice period that may be awkward to change later, one option would be to add an admin function that allows the deployer to change the parameter on the live contract without having to redeploy. However, I would highlight a couple of caveats to this approach:
1. Unrestricted control of the wait period would give the admin power to effectively rug funds by setting the wait to a very large value. That in turn places a higher target value on the admin keys, e.g. for social engineering attacks. To protect depositors, the admin function ought therefore to have some safeguards that restrict the admin's power, for example:
   - A hardcoded maximum value (30 days?)
   - An enforced wait period before admin-triggered changes to the value come into effect, so that depositors not happy with the new configuration have time to trigger exits or withdrawals.
2. Implementing such a function does not save us from the need to set up processes for monitoring and deciding when the parameter should be updated. The situation in which it does buy us something is if we do decide there needs to be a retuning and there are no other upcoming updates to the stake registry that would necessitate a redeployment anyway.
3. We need to decide whether decisions about calling the update function necessitate a SWIP or if it can be conducted more efficiently through some other sufficiently well-defined process.

Understanding these factors, especially (1), means that trying to get this function into the current SWIP could further delay the decision process. My view is that it isn't really necessary right now for the following reasons:
1. We are probably going to need to update the stake registry again, and further parameter updates can be bundled with that change.
2. The choice between 14 or 28 days, or even 7 days, probably does not matter all that much at the current network size. The real win here is coming from bringing it down from ∞ to less than a month.

Nonetheless, if you want it I also don't mind adding it.

awmacpherson:
Posting expanded rationale as a reply first so it's easier to discuss. If it satisfies you @significance, I'll merge the arguments into the SWIP.

## TLDR

Reducing the withdrawal notice period attracts more stake but carries more risk of inducing operator behaviour that leads to a loss of service — call this *network risk.* Identifying the sweet spot depends on being able to measure this risk.

The current proposal text calls for a withdrawal notice period of 28 days. Other contributors have suggested that the wait could be shorter: either 7 or 14 days. Which should we choose?

**Assessment.** We expect that reducing the notice period from 28 days to 7 days (or any value between 7 and 28 days) will not attract a significant quantity of additional stake or nodes; nor will it significantly decrease the risk of dirty exits. On the other hand, the excess network risk associated with such reduction is hard to quantify: more research is needed. That said, we don't know any *a priori* reason to think it will be critical.

**Recommendation.** Dedicate some resources to identifying and monitoring risk signals that could inform our choice of notice period. However, do not block the update on the outcome of such research — we have enough information to agree on a value now and move forward.

We suggest starting at the upper limit of the discussed range, i.e. 28 days, because this has the lowest risk, and we are not yet well equipped to measure how much additional risk would be added by a shorter period. 

## Analysis

Our proposals for stake withdrawal and exit notice periods argue that the risk of service issues is greater for shorter notice periods because of the following two effects:

* *Filtering.* A notice period filters out some operator policies that involve withdrawing stake and potentially also turning off nodes. Specifically, if the wait period is $N$ days, policies that require the operator to withdraw stake less than $N$ days after observing a signal are not possible. Increasing $N$ enlarges the family of policies that are filtered out.

* *Signalling.* Broadcast the operator intent to reduce service in advance of the change taking effect, allowing other ecosystem participants to react before the service reduction occurs. 

  * *Preemptive repair.* Advance signalling gives other operators a window in which to repair the service gap that will be left by the exiting nodes before such a gap ever appears. Increasing $N$ enlarges this window, hence increasing the chances that preemptive repair occurs.

  * *Price signalling.* Storage prices are a function of supply and demand signals available onchain. If there is a function of data available onchain $N$ days in advance that can be used to bound prices above, this makes prices more predictable, improving DX (compare https://github.com/ethersphere/SWIPs/issues/86).

    Concretely, in Swarm today, the supply side input to the pricing function is the number of nodes that reveal in each neighbourhood. If there are no dirty exits and all staked nodes reveal each round, a notice period of $N$ days for exits allows us to bound below the number of nodes in each neighbourhood $N$ days into the future. This gives us an upper bound on pricing (assuming fixed demand).

    This reasoning survives some changes to the way prices are quoted in Swarm: as long as the price depends on the number of nodes revealing in each neighbourhood is an input, advance signalling of decreases in these numbers improves price predictability.

There is also a risk effect that increases with *longer* notice periods:

* *Dirty exit.* If the exit wait is too long, operators may choose to turn off nodes without respecting the withdrawal notice period, forgoing expected revenue to which they would otherwise be entitled. This risk increases as $N$ increases, with the extreme being the status quo $N=\infty$ in which only dirty exits are possible. Dirty exit risk could likely be counteracted by introducing explicit penalties for missed rounds.

awmacpherson:
Just to make things a bit more concrete, here are some candidates for risk signals that we can monitor to evaluate our choice of wait period:

* *Worst case service quality.* Smallest number of staked nodes in any neighbourhood.
* *Reaction to market signals.* Monitor the relationship at various lags of operator behaviour, especially node exits, to market signals such as the price of BZZ or Swarm network revenue.
* *Repair events.* For each replication rate $N\leq 4$, monitor events in which the number of staked nodes in a given neighbourhood drops below $N$. For each such event, record the *time to repair (TTR)*, that is, the length of time before the node count in that neighbourhood returns to $N$.
* *Preemptive repair events.* As above, but measure the exit time from when the exit is *enqueued* rather than executed. Identify events in which repair occurs — i.e. the node count increases — before the exit is executed.
* *Dirty exits.* Monitor nodes that trigger an exit but miss reveals before the wait period has elapsed.

significance:
yes, sounds good, thank you for this @awmacpherson 

i see no reason that it should be shorter necessarily and would be happy to start with 28 days and modify this based on signals. i note that contract admin currently has the ability to enable withdrawal immediately and as such enabling a more granular but monotonically decreasing means of adjusting this would be congruent with the current permissions schedule and as such uncontroversial as far as i can see

PR #98, opened 2026-06-24
SWIP 049 - Fixed set of stamps for stamp sampling phase

lat-murmeldjur:
Bee and the Redistribution contract must use the same set of batches and indexes. This SWIP fixes that set when sampling begins and keeps it fixed through the last claim block. A batch may be used only if it already existed, had at least 456 blocks of balance at the price in force when sampling began, and already contained the proved index at that time. New batches and newly created indexes remain usable for uploads immediately, but they cannot be used by an already open redistribution round.

Furthermore, the suggestion is for the PostageStamp contract to keep one previous price and two previous batch depths. The Redistribution contract can use that history to reconstruct the price and index range that applied when sampling began at claim. A top-up is allowed only while the batch still has at least six rounds of balance, and a dilution must leave at least six rounds of balance. The claim verifies that the batch is still present and live, verifies all postage proofs, including that the index existed at the start of sampling, and only then allows the price to change.

### DIFF
--- a/SWIPs/swip-49.md
+++ b/SWIPs/swip-49.md
@@ -1,0 +1,737 @@
+---
+swip: SWIP 049
+title: Fixed postage-stamp usability for redistribution rounds
+author: lat-murmeldjur
+status: Draft
+type: Standards Track
+category: Core
+created: 2026-06-24
+---
+
+# Fixed postage-stamp usability for redistribution rounds
+
+## Abstract
+
+Phase 5 adds a postage-stamp sample beside the existing chunk-data sample. For each distinct stamp identity, Bee calculates `hash(batchId, fullStampIndex, roundAnchor)` and keeps the required lowest values. One purchased index can therefore contribute only one transformed value in a round, so overreporting storage depth requires a correspondingly large number of real stamp indexes. The separate chunk sample remains necessary to show that participants also store the same chunk data.
+
+For the two samples to be usable for claim, Bee and the Redistribution contract must use the same set of batches and indexes. This SWIP fixes that set when sampling begins and keeps it fixed through the last claim block. A batch may be used only if it already existed, had at least 456 blocks of balance at the price in force when sampling began, and already contained the proved index at that time. New batches and newly created indexes remain usable for uploads immediately, but they cannot be used by an already open redistribution round.
+
+Furthermore, the suggestion is for the PostageStamp contract to keep one previous price and two previous batch depths. The Redistribution contract can use that history to reconstruct the price and index range that applied when sampling began at claim. A top-up is allowed only while the batch still has at least six rounds of balance, and a dilution must leave at least six rounds of balance. The claim verifies that the batch is still present and live, verifies all postage proofs, including that the index existed at the start of sampling, and only then allows the price to change.
+
+## Specification
+
+### When the usable batch and index set is fixed
+
+A redistribution round contains 152 blocks:
+
+- 38 commit blocks;
+- 38 reveal blocks; and
+- 76 claim blocks.
+
+For target round `r`, sampling begins at the first reveal block of round `r - 1`. This is the first scheduled block in which a successful reveal can produce the anchor used by target round `r`. If the preceding reveal phase has no successful reveal, the existing skipped-round seed derivation is used and the target round remains playable. The sampling boundary does not move.
+
+The sampling-start block is:
+
+```text
+samplingStartBlock(r) = (r - 1) * 152 + 38
+```
+
+The state used by the round is the state after the preceding block has been processed. A batch creation or dilution included before `samplingStartBlock` may therefore affect the target round. An operation included in `samplingStartBlock` or later is too late for that round.
+
+Counting the sampling-start block and the final target-round claim block, the round's sampling-to-claim window spans 266 blocks. During that window, later postage operations must not change:
+
+- whether a batch may be used by the target round; or
+- how many indexes from that batch may be used by the target round.
+
+### When a batch and stamp index may be used
+
+A stamp index may be used by target round `r` only when all of the following are true:
+
+1. The batch was created before `samplingStartBlock(r)`.
+2. At `samplingStartBlock(r)`, the batch had enough remaining balance for 456 blocks at the price then in force.
+3. The full stamp index fitted within the batch depth that existed before `samplingStartBlock(r)`.
+4. The batch is still present and live when the target claim verifies the stamp.
+5. The existing signature, bucket-alignment, inclusion-proof, and other postage checks pass.
+
+The target claim can be submitted as late as the end of the 266-block window. Requiring 456 blocks of balance at sampling start leaves 190 starting-price blocks of headroom over that claim window. This headroom is used to tolerate the price increase that the preceding round's claim may apply while target-round sampling is already in progress.
+
+### Recovering the price and balance at sampling start
+
+`PostageStamp` already stores:
+
+- `lastPrice`, the price currently in force;
+- `lastUpdatedBlock`, the block in which that price became active; and
+- `totalOutPayment`, cumulative per-chunk outpayment at that update block.
+
+This SWIP adds:
+
+```text
+previousPrice
+```
+
+`previousPrice` is the price that was in force immediately before `lastPrice`.
+
+After activation, `PostageStamp.setPrice` is called only by a redistribution claim. The target claim verifies all postage proofs before applying its own price update. Therefore, when target-round postage proofs are checked, at most one price update can have occurred after sampling began: the update made by the preceding round's claim.
+
+The price and cumulative outpayment at sampling start are reconstructed as follows:
+
+```text
+if lastUpdatedBlock < samplingStartBlock:
+    priceAtStart = lastPrice
+    outPaymentAtStart =
+        totalOutPayment
+        + (samplingStartBlock - lastUpdatedBlock) * lastPrice
+else:
+    priceAtStart = previousPrice
+    outPaymentAtStart =
+        totalOutPayment
+        - (lastUpdatedBlock - samplingStartBlock) * previousPrice
+```
+
+The fixed balance threshold for the target round is:
+
+```text
+requiredNormalisedBalance =
+    outPaymentAtStart + 456 * priceAtStart
+```
+
+Bee and the Redistribution contract use this same threshold. A later price update does not change it.
+
+### Minimum balance required for batch operations
+
+Every operation that creates a batch or reduces its balance per chunk must respect a minimum of six complete redistribution rounds at the price in force when the operation executes. Six rounds are:
+
+```text
+6 * 152 = 912 blocks
+```
+
+The rules are:
+
+- a new or imported batch must start with at least `912 * lastPrice` per chunk;
+- a batch may be topped up only if it still has at least `912 * lastPrice` per chunk before the top-up; and
+- a dilution is allowed only if the resulting per-chunk balance is at least `912 * lastPrice`.
+
+A deployment may require a larger minimum, but it must not permit less than 912 blocks.
+
+These rules remove the need for per-batch balance history. A top-up cannot rescue a nearly expired batch into an already open round, because the batch must already have six rounds of balance before it can be topped up. A dilution cannot remove a batch from an already open round, because the diluted batch must still have six rounds of balance afterward.
+
+New batches and the additional indexes created by dilution remain usable for uploads immediately. The introduction block only determines the first redistribution round in which they may be sampled and used for claim.
+
+### Recovering the depth that existed at sampling start
+
+The live `Batch` continues to store its current depth and `lastUpdatedBlockNumber`. For this purpose, `lastUpdatedBlockNumber` is the block in which the current depth was introduced. Top-ups do not change it.
+
+Each batch additionally stores the two depths immediately before the current depth, together with the blocks in which those depths were introduced:
+
+```text
+PreviousDepths {
+    previousDepth
+    previousDepthBlock
+    olderDepth
+    olderDepthBlock
+}
+```
+
+For a target round, the usable depth is the newest recorded depth whose introduction block is earlier than `samplingStartBlock`. The contract checks the current depth first, then the previous depth, then the older depth. If none of them existed before sampling began, the batch did not yet exist for that target round.
+
+Two preceding depths are required because neighboring sampling-to-claim windows overlap. A later round can fix its batch scope while the preceding target round can still be claimed.
+
+For example:
+
+```text
+Before round r starts sampling:
+    depth = 20
+
+After round r starts sampling, but before round r+1 starts sampling:
+    dilute to 21
+    dilute again to 22
+
+After round r+1 starts sampling, while round r is still claimable:
+    dilute to 23
+```
+
+The required result is:
+
+```text
+round r uses depth 20
+round r+1 uses depth 22
+uploads and later rounds use depth 23
+```
+
+The contract must therefore retain:
+
+```text
+older depth    = 20
+previous depth = 22
+current depth  = 23
+```
+
+Depth 21 does not need to be retained because no sampling boundary observed it. The first dilution after a sampling boundary rotates the depth history once. Further dilutions before the next sampling boundary replace only the current depth. This permits multiple dilutions without losing a depth that an open target claim still needs.
+
+### Target-claim verification and price-update order
+
+The target round number determines `samplingStartBlock`. At the beginning of claim verification, the Redistribution contract reconstructs `requiredNormalisedBalance` once and uses the same value for every postage proof in the claim. This includes postage attached to the chunk-data sample and postage proved for the separate stamp sample.
+
+For each proved stamp, the PostageStamp contract supplies:
+
+- the current batch owner;
+- the immutable bucket depth;
+- the current normalised balance; and
+- the historical batch depth that existed before sampling began.
+
+The Redistribution contract accepts the stamp only if:
+
+- a historical depth exists for the target sampling start;
+- the batch's normalised balance is at least `requiredNormalisedBalance`;
+- the batch is still live under `currentTotalOutPayment()`;
+- the proved index already existed within the batch's index range when sampling began; and
+- the existing bucket and signature checks pass.
+
+The fourth check is the claim-time proof that a later dilution did not create the
+proved index. The contract first resolves `depthAtSamplingStart`, extracts the
+index-within-bucket from `fullStampIndex`, and checks:
+
+```text
+getPostageIndex(fullStampIndex)
+    < postageStampIndexCount(depthAtSamplingStart, bucketDepth)
+```
+
+For example, an index can be valid under the batch's larger live depth at claim
+time but still be rejected because it was outside the smaller index range that
+existed when sampling began. Claim verification always uses the latter range.
+
+An expired batch is never accepted. If expiry cleanup has already run, the batch is absent from the normal batch mapping. If cleanup has not yet run, the explicit live-balance check rejects it. No expired-batch tombstone or secondary claim registry is introduced.
+
+`winnerSelection` must not update the postage price. It returns the redundancy count to `claim`. The claim then verifies every chunk proof and postage proof. After all proofs have succeeded, it marks the round claimed, calls `PriceOracle.adjustPrice`, and continues with the existing withdrawal. Because `PostageStamp.setPrice` accounts up to the current block at the old price, this target-round price update is not retroactive and cannot invalidate the proofs just checked.
+
+The stamp transformation is exactly:
+
+```text
+hash(batchId, fullStampIndex, roundAnchor)
+```
+
+### Tolerance of skipped-round price catch-up
+
+The 456-block requirement is intended to keep a sampled batch in the normal batch registry for the complete 266-block sampling-to-claim window.
+
+The earliest preceding-round claim can update the price 38 blocks after target sampling begins. A batch admitted with exactly 456 starting-price blocks has then spent 38 blocks and still holds 418 starting-price block units.
+
+Under the current PriceOracle, the largest ordinary increase for the claimed round is:
+
+```text
+1049206 / 1048576
+```
+
+Each skipped round compounds the maximum increase:
+
+```text
+1049417 / 1048576
+```
+
+After 755 skipped rounds, the combined maximum price factor is approximately:
+
+```text
+1.832897789
+```
+
+This is an increase of approximately **83.2898%**. The remaining 418 starting-price block units then represent approximately:
+
+```text
+418 / 1.832897789 = 228.054 blocks
+```
+
+Using the conservative 266-block window, 228 blocks remain after the earliest possible price update. The batch therefore remains live through the latest target-round claim.
+
+After 756 skipped rounds, the remaining balance represents approximately 227.871 blocks at the increased price, so survival through the complete conservative window is no longer guaranteed. Integer division in the oracle can only reduce the realised increase, so 755 skipped rounds is a conservative guaranteed tolerance under the current constants.
+
+This number is a safety analysis, not an on-chain limit. If a larger catch-up causes a selected batch to expire before proof verification, the target claim rejects that batch normally. No expired-batch data is retained for redistribution.
+
+### Bee behaviour
+
+Before constructing the samples for target round `r`, Bee processes canonical postage events through the block immediately before `samplingStartBlock(r)`. It then applies the same rules as the contracts:
+
+- the batch and applicable depth must have been introduced before sampling began;
+- the batch must meet the 456-block threshold calculated from the price and cumulative outpayment at sampling start; and
+- only indexes within that historical depth may be sampled.
+
+Bee may reconstruct this state locally from events. The PostageStamp read functions defined by this SWIP are the canonical contract reference. Later creation, top-up, dilution, expiry, or price updates do not change an already constructed target-round sample.
+
+### Batch ID reuse
+
+Expiry continues to remove a batch from the ordered tree and delete its live batch data and depth history. The batch ID remains permanently consumed. Recreating an expired ID could make old signatures usable under a new batch incarnation, so creation and import paths must reject every previously used batch ID.
+
+## Rationale
+
+The design adds one global historical price and two historical depths per batch. It does not add per-round price snapshots, per-batch balance checkpoints, or retained expired batches.
+
+One previous price is sufficient because claims are the only price-update path and the target claim updates the price only after proof verification. Two previous depths are sufficient because only two neighboring sampling-to-claim windows overlap under the current 152-block schedule. The six-round operation minimum prevents later top-ups and dilutions from crossing the 456-block usability boundary fixed for an open round.
+
+## Security considerations
+
+The fixed scope prevents a participant from learning a favorable anchor and then buying a batch or diluting an existing batch to create indexes for that target round.
+
+The pre-top-up check prevents a nearly expired batch from being rescued into an open round. The post-dilution check prevents a usable batch from being forced out by dilution. Historical depth prevents later dilution from expanding the index range used by the claim. Historical price and reordered claim execution ensure that the preceding round's price update does not change the balance criterion used by Bee and the target claim.
+
+The 755-skipped-round figure depends on the current phase lengths and PriceOracle constants. A future change to round length, adjustment rates, or price-update ordering must re-evaluate the 456-block and 912-block parameters.
+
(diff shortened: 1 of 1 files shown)


SWIPs/swip-49.md
```diff
@@ -0,0 +1,737 @@
...
+type: Standards Track
+category: Core
+created: 2026-06-24
+---
+
+# Fixed postage-stamp usability for redistribution rounds
+
+## Abstract
+
+Phase 5 adds a postage-stamp sample beside the existing chunk-data sample. For each distinct stamp identity, Bee calculates `hash(batchId, fullStampIndex, roundAnchor)` and keeps the required lowest values. One purchased index can therefore contribute only one transformed value in a round, so overreporting storage depth requires a correspondingly large number of real stamp indexes. The separate chunk sample remains necessary to show that participants also store the same chunk data.
```
zelig, line 15: the "transformed hash" is using the `roundAnchor` as a prefix, so it is `hash( roundAnchor, batchId, index)` to be the same as the phase 4 transformed hash.

SWIPs/swip-49.md
```diff
@@ -0,0 +1,737 @@
...
+
+### When the usable batch and index set is fixed
+
+A redistribution round contains 152 blocks:
+
+- 38 commit blocks;
+- 38 reveal blocks; and
+- 76 claim blocks.
+
+For target round `r`, sampling begins at the first reveal block of round `r - 1`. This is the first scheduled block in which a successful reveal can produce the anchor used by target round `r`. If the preceding reveal phase has no successful reveal, the existing skipped-round seed derivation is used and the target round remains playable. The sampling boundary does not move.
```
zelig, line 31: > For target round `r`, sampling begins at the first reveal block of round `r - 1`.

this is not true, but as you say: 

>  This is the first scheduled block in which a successful reveal can produce the anchor used by target round `r`.

I suggest a dofferent wording:

The anchor target round `r` is set by the first successful reveal transaction of the previous round. Since the earliest block this can happen is at the first reveal block of the previous round, we can set that as the "starting block" for sampling.

SWIPs/swip-49.md
```diff
@@ -0,0 +1,737 @@
...
+
+The sampling-start block is:
+
+```text
+samplingStartBlock(r) = (r - 1) * 152 + 38
+```
+
+The state used by the round is the state after the preceding block has been processed. A batch creation or dilution included before `samplingStartBlock` may therefore affect the target round. An operation included in `samplingStartBlock` or later is too late for that round.
+
+Counting the sampling-start block and the final target-round claim block, the round's sampling-to-claim window spans 266 blocks. During that window, later postage operations must not change:
```
zelig, line 41: 266 = 2 * 152 - 38

SWIPs/swip-49.md
```diff
@@ -0,0 +1,737 @@
...
+
+A stamp index may be used by target round `r` only when all of the following are true:
+
+1. The batch was created before `samplingStartBlock(r)`.
+2. At `samplingStartBlock(r)`, the batch had enough remaining balance for 456 blocks at the price then in force.
+3. The full stamp index fitted within the batch depth that existed before `samplingStartBlock(r)`.
+4. The batch is still present and live when the target claim verifies the stamp.
+5. The existing signature, bucket-alignment, inclusion-proof, and other postage checks pass.
+
+The target claim can be submitted as late as the end of the 266-block window. Requiring 456 blocks of balance at sampling start leaves 190 starting-price blocks of headroom over that claim window. This headroom is used to tolerate the price increase that the preceding round's claim may apply while target-round sampling is already in progress.
```
zelig, line 56: how do you know that this length tolerates the potential price increase of the previous round's claim. This must be calculated.

SWIPs/swip-49.md
```diff
@@ -0,0 +1,737 @@
...
+
+The rules are:
+
+- a new or imported batch must start with at least `912 * lastPrice` per chunk;
+- a batch may be topped up only if it still has at least `912 * lastPrice` per chunk before the top-up; and
+- a dilution is allowed only if the resulting per-chunk balance is at least `912 * lastPrice`.
+
+A deployment may require a larger minimum, but it must not permit less than 912 blocks.
+
+These rules remove the need for per-batch balance history. A top-up cannot rescue a nearly expired batch into an already open round, because the batch must already have six rounds of balance before it can be topped up. A dilution cannot remove a batch from an already open round, because the diluted batch must still have six rounds of balance afterward.
```
zelig, line 116: why 6? how do you make sure that the degree of the change of price has these properties?

SWIPs/swip-49.md
```diff
@@ -0,0 +1,737 @@
...
+- the current normalised balance; and
+- the historical batch depth that existed before sampling began.
+
+The Redistribution contract accepts the stamp only if:
+
+- a historical depth exists for the target sampling start;
+- the batch's normalised balance is at least `requiredNormalisedBalance`;
+- the batch is still live under `currentTotalOutPayment()`;
+- the proved index already existed within the batch's index range when sampling began; and
+- the existing bucket and signature checks pass.
```
zelig, line 188: can we just switch the last two and elaborate the last?

SWIPs/swip-49.md
```diff
@@ -0,0 +1,737 @@
...
+
+The fourth check is the claim-time proof that a later dilution did not create the
```
zelig, line 190: The last check...

SWIPs/swip-49.md
```diff
@@ -0,0 +1,737 @@
...
+    external view returns (address owner, uint8 depthAtSamplingStart, uint8 bucketDepth, uint256 normalisedBalance);
+```
+
+## Appendix B: `Redistribution` changes relative to the Phase 5 starting diff
+
+The Phase 5 diff already introduces the stamp commitment and proofs, including
+the transformed-stamp encoding:
+
+```solidity
+keccak256(abi.encodePacked(postageId, index, roundAnchor))
```
zelig, line 604: roundAnchor in front again

SWIPs/swip-49.md
```diff
@@ -0,0 +1,737 @@
...
+    // This is the claim-time proof that the index already existed when
+    // sampling began. A later dilution may make a larger index valid under the
+    // live depth, but it cannot make that index valid for this target round.
+    uint256 indexWithinBucket = getPostageIndex(fullStampIndex);
+    uint256 indexCountAtSamplingStart =
+        postageStampIndexCount(depthAtSamplingStart, bucketDepth);
+    if (indexWithinBucket >= indexCountAtSamplingStart) {
+        revert IndexOutsideSet(batchId);
+    }
+    if (getPostageBucket(fullStampIndex) != addressToBucket(chunkAddress, bucketDepth)) {
```
zelig, line 697: Important: his is the alignment criterion on valid stamps. It seems you also missed the retrievability criterion which requires stamps with chunk address in the claimant's neighbourhood, i.e., match the node overlay on the first depth bits. Or is it there somewhere?

Note that all these ultimately match also on the roundAnchor is the result of the responsiblity criterion (ie that the node is in the nh, the roundAnchor matches the overlay on depth bits) which is already checked at reveal

SWIPs/swip-49.md
```diff
@@ -0,0 +1,737 @@
...
+}
+
+function redistributionBatchAt(bytes32 batchId, uint256 samplingStartBlock)
+    external
+    view
+    returns (address owner, uint8 depthAtSamplingStart, uint8 bucketDepth, uint256 normalisedBalance)
+{
+    Batch storage batch = batches[batchId];
+    if (batch.owner == address(0) || batch.normalisedBalance <= currentTotalOutPayment()) {
+        revert BatchNotUsableForRedistribution(batchId);
```
zelig, line 576: ` <= currentTotalOutPayment())` 

are you sure?

SWIPs/swip-49.md
```diff
@@ -0,0 +1,737 @@
...
+function setMinimumValidityBlocks(uint64 _value) external {
+    if (!hasRole(DEFAULT_ADMIN_ROLE, msg.sender)) revert AdministratorOnly();
+    if (_value < MIN_OPERATION_VALIDITY_BLOCKS) revert MinimumValidityTooShort();
+    minimumValidityBlocks = _value;
+}
+```
+
+After activation, `setPrice` is reached only through a redistribution claim.
+
+5. **Delete depth history on expiry, but do not release the batch ID.**
```
zelig, line 508: not release (delete?) the batchID? why not?

SWIPs/swip-49.md
```diff
@@ -0,0 +1,737 @@
...
+- the batch must meet the 456-block threshold calculated from the price and cumulative outpayment at sampling start; and
+- only indexes within that historical depth may be sampled.
+
+Bee may reconstruct this state locally from events. The PostageStamp read functions defined by this SWIP are the canonical contract reference. Later creation, top-up, dilution, expiry, or price updates do not change an already constructed target-round sample.
+
+### Batch ID reuse
+
+Expiry continues to remove a batch from the ordered tree and delete its live batch data and depth history. The batch ID remains permanently consumed. Recreating an expired ID could make old signatures usable under a new batch incarnation, so creation and import paths must reject every previously used batch ID.
+
+## Rationale
```
zelig, line 263: This subsection should be much mich more verbose. Please explain even if briefly the current situation, the vulnerability and the fix. Own up to an honest scope and let us in on limitations.

SWIPs/swip-49.md
```diff
@@ -0,0 +1,737 @@
...
+
+### Bee behaviour
+
+Before constructing the samples for target round `r`, Bee processes canonical postage events through the block immediately before `samplingStartBlock(r)`. It then applies the same rules as the contracts:
+
+- the batch and applicable depth must have been introduced before sampling began;
+- the batch must meet the 456-block threshold calculated from the price and cumulative outpayment at sampling start; and
+- only indexes within that historical depth may be sampled.
+
+Bee may reconstruct this state locally from events. The PostageStamp read functions defined by this SWIP are the canonical contract reference. Later creation, top-up, dilution, expiry, or price updates do not change an already constructed target-round sample.
```
zelig, line 257: may? you mean Bee must be able to and in fact is able to , because ?

SWIPs/swip-49.md
```diff
@@ -0,0 +1,737 @@
...
+
+### Batch ID reuse
```
zelig, line 259: This is a totally unrelated issue that is kind of hidden here. Consider reifing it as a separate SWIP

SWIPs/swip-49.md
```diff
@@ -0,0 +1,737 @@
...
+
+```text
+418 / 1.832897789 = 228.054 blocks
+```
+
+Using the conservative 266-block window, 228 blocks remain after the earliest possible price update. The batch therefore remains live through the latest target-round claim.
+
+After 756 skipped rounds, the remaining balance represents approximately 227.871 blocks at the increased price, so survival through the complete conservative window is no longer guaranteed. Integer division in the oracle can only reduce the realised increase, so 755 skipped rounds is a conservative guaranteed tolerance under the current constants.
+
+This number is a safety analysis, not an on-chain limit. If a larger catch-up causes a selected batch to expire before proof verification, the target claim rejects that batch normally. No expired-batch data is retained for redistribution.
```
zelig, line 247: I do not understand these two sentences

SWIPs/swip-49.md
```diff
@@ -0,0 +1,737 @@
...
+1049417 / 1048576
+```
+
+After 755 skipped rounds, the combined maximum price factor is approximately:
+
+```text
+1.832897789
+```
+
+This is an increase of approximately **83.2898%**. The remaining 418 starting-price block units then represent approximately:
```
zelig, line 237: 'represent' is a funny word to use here. I think it is better to say 'covers the storage rent for a shorter period with the higher price.

SWIPs/swip-49.md
```diff
@@ -0,0 +1,737 @@
...
+1049206 / 1048576
+```
+
+Each skipped round compounds the maximum increase:
+
+```text
+1049417 / 1048576
+```
+
+After 755 skipped rounds, the combined maximum price factor is approximately:
```
zelig, line 231: No no no. You seem to be compounding a multiplicative factor. But it is actually just the price itself.
No matter how many skipped rounds happened. the  rate of change is capped at a maximum of the above ratio. which means that the price change is minuscule during a

UPDATE: I was wrong herein the sense that surely the multiplicative factor applies per round, In the caes of a skipped round their application cannot happen, so the cumulated change does apply at once right after the stretch of skipped rounds, in which there is no claim transactions.

SWIPs/swip-49.md
```diff
@@ -0,0 +1,737 @@
...
+418 / 1.832897789 = 228.054 blocks
+```
+
+Using the conservative 266-block window, 228 blocks remain after the earliest possible price update. The batch therefore remains live through the latest target-round claim.
+
+After 756 skipped rounds, the remaining balance represents approximately 227.871 blocks at the increased price, so survival through the complete conservative window is no longer guaranteed. Integer division in the oracle can only reduce the realised increase, so 755 skipped rounds is a conservative guaranteed tolerance under the current constants.
+
+This number is a safety analysis, not an on-chain limit. If a larger catch-up causes a selected batch to expire before proof verification, the target claim rejects that batch normally. No expired-batch data is retained for redistribution.
+
+### Bee behaviour
```
zelig, line 249: lets not call it `bee` but node client changes instead. There are a lot of alternative clients

zelig (changes requested):
UPDATE: 
After talking to the author I am conviced that this is close to the right solution. The SWIP is difficult though.
So i think it would help if the text was iterated on:

- add proper context, motivation
- make it clear that this is about the storage incentives phase 5 problem
- makt it clear which aspect of the phase 5 this solves and which it leaves open
- the calculations for the validity periods should be either before the 'spec' or referenced.
- the calculations should be done formally using generics first such as 'maximum per-round-price-increase' and  concrete numbers  only after, emphasize that skipped rounds is important here NOT because it is the highest increase but because it causes the cumulation of price increase without being applied due to lack of transactions.
- I recommend better naming (of functions and variables)
- align transformed slot reference with phase 4 ones (anchor nonce as a prefix)


I have read through it and commented thoroughly, but
there is a major flaw here https://github.com/ethersphere/SWIPs/pull/98/changes#r3485677904
and I am not sure of the motivation for increased complexity.

What we do need is a delay in dilution and (maybe require a long effective ttl after)

This is a short and Sweet SWIP of its own, not worth mixing in othr stuff
https://github.com/ethersphere/SWIPs/pull/98/changes#r3484124428

zelig:
Right. I  talked it through with the author and they conviced me that the proposed solution is necessary, sound and very likely not possible to simplify much.

SWIPs/swip-49.md
zelig: Maybe it is worth elaboratign why only the change cumulated in the previous rounds matter and not the 2 price changes after.

SWIPs/swip-49.md
lat-murmeldjur: Yes so this is the actual aim. The claim will only check whether the balance was at least 456 blocks remaining by the price in force at the time of the starting block for sampling, which is an exact balance requirement that does not shift after sampling start block. However, the batch has to remain in the registry and not expire despite the price change that happens after the starting block of sampling. So this 456 blocks is chosen so that it tolerates an 80% price increase and makes sure that a batch that had only 456 blocks of validity left will still not expire and be deleted from the registry after 266 blocks (which is the number of blocks between the last possible claim and the sampling start block). Will add this explanation to the SWIP too

SWIPs/swip-49.md
lat-murmeldjur: 6 rounds was chosen arbitrarily. It could be 4, or even close to 3, but not less, so that it's not possible to top up a batch that has less remaining balance than the 3 rounds worth of balance required by the sampling, or that it's not possible to dilute a batch to less balance than what is required by the sampling.

What needs to be considered here is potential price decrease caused by a claim. It should not be possible to decrease a batch balance by dilution to become less than the 456 block worth of remaining balance by the previous price even after a price decrease. So if the maximum price decrease is 1%, the minimum balance that can be created by batch operations needs to be around 1% more than 456 blocks. Also going to add this to the SWIP description

SWIPs/swip-49.md
lat-murmeldjur: Batch ID reuse is really not necessary to be part of this SWIP. If a batch is recreated, it will have new creation blocks, and will not be usable for sampling in the same round anyway

SWIPs/swip-49.md
lat-murmeldjur: If there are for example 800 skipped rounds,  the price increase catch-up could be more than 80%, and it could mean that a batch that was sampled with near minimal validity could become expired by the last claim block. However this tolerance is already huge and only influences rare edge cases, a batch that has a bit more than the minimum balance would already survive more than 80% price increase too.

lat-murmeldjur:
Thank you for the review

So to state the reasoning behind the necessity of the different parts behind this SWIP:

Choosing 456 blocks of validity based on the price in force at the sampling start block.
This seems reasonable, because waiting for the claim to actually set a price would decrease the time for sampling.
So to have a definitive minimal balance that won't shift until the claim it seems reasonable to use the price in effect when the sampling can start earliest. For  this to work we need some tolerance expecting a potential price increase (456 blocks of validity instead of 266), and we need to remember the price in effect at sampling start, so that the claim can verify each batch meets this exact requirement. This creates an unambigous balance requirement for both the client and the contract with the introduction of one new contract variable and a short formula function to calculate this balance requirement. 

Choosing 912 block of minimum validity accepted for the result of batch operations:
Could be decreased to little more than 456 blocks. Otherwise this makes sure that batch operations can not move a batch in or out of the scope of valid batches for a round that already began sampling. 

Dilution depth history. 
Whenever a dilution happens, there can be two rounds ongoing, one that is close to the claim, and the other that just started sampling. For both of these rounds we need to remember the specific depth of a batch before the sampling of the round began, so that new indexes can not be created during sampling and used already in the same period claim (however they still remain immediately usable for uploads). So at any given moment, we might need to remember the depth of a batch before the latest sampling began, and before the sampling for the preceding round began 156 blocks before. This SWIP is really close to minimising the administrative cost of keeping track of this, as this proposal tracks exactly up to 2 previous depths for any batch (with their introduction blocks). It does not require extra updates, only when a dilution happens, and in claim we can deduct the range of valid indexes for a batch from this limited history for any batch. The achievement of this solution is that dilutions can still happen with any frequency, the UX remains unchanged and not constrained compared to the previous experience.

Batch ID reuse 
Not necessary

0xCardiE:
Appendix B is a diff against a Phase 5 branch that doesn't exist or I am not aware of where this is. 
Was expecting that this SWIP is based on current phase 4 branch + swip 51 code

0xCardiE:
### `setPrice` is not claim-only, so the one-update assumption doesn't hold

The rollback in `_priceStateAtSamplingStart` is sound only if at most one price update
happened since `samplingStartBlock`. The SWIP gets that from "after activation,
`PostageStamp.setPrice` is called only by a redistribution claim".

That premise is false today. `PriceOracle.setPrice(uint32)` (`src/PriceOracle.sol:76`) is
gated on `DEFAULT_ADMIN_ROLE` alone — no round limit, no rate limit — and it calls
`PostageStamp.setPrice`. A single admin call during an open round is a second update.

Consequence: the rollback charges the whole interval at `previousPrice`, so
`outPaymentAtStart` and the resulting threshold are wrong. No revert — the claim silently
evaluates a different batch set than Bee did, which is the exact divergence this SWIP exists
to prevent. Whoever holds the oracle admin key can move that set in either direction.

Security Considerations doesn't mention the admin path at all.

**Ask:** either restrict `PriceOracle.setPrice` post-activation, or store
`previousPriceUpdatedBlock` and revert when it is `> samplingStartBlock` so the precondition
is checked rather than assumed. (I've implemented the latter.)

0xCardiE:
What needs deciding is whether SWIP-049's new import minimum is supposed to apply to migration.  because as written it means a migration silently discards every batch with under ~24h of runway. Which could be ok, just that we explicitly define this as expected.

0xCardiE:
### The Tolerance section's arithmetic doesn't survive SWIP-050

I verified the derivation against the current schedule and it reproduces exactly: 266-block
window, earliest price update at `samplingStart+38`, 418 of 456 units left covering 228
blocks, factor 1.833, 755 skipped rounds. No complaints about 456 itself.

But step one  "the earliest preceding-round claim can update the price 38 blocks after target
sampling begins"  depends on the claim phase opening at `B+76`. Under SWIP-050 it opens at
`B+133`, i.e. `samplingStart+95`. Rerunning: 361 units covering 171 blocks, factor 2.111,
~931 skipped rounds.

456 stays safe  it gets *more* conservative, so no parameter changes. The issue is that the
published derivation is wrong once the companion SWIP lands, and this section states the rule
itself: "a future change to round length, adjustment rates, or price-update ordering must
re-evaluate the 456-block and 912-block parameters." SWIP-050 is that change.

0xCardiE:
### Two small things in Appendix A item 6

**1. The two new views disagree on future blocks.** `_priceStateAtSamplingStart` reverts
`FutureSamplingStartBlock`; `redistributionBatchAt` has no such guard, so a future block takes
the `batch.lastUpdatedBlockNumber < samplingStartBlock` branch and returns the *current* depth
as though it were historical.

Unreachable on-chain  `claim` derives the round from `block.number`  but these are the
Bee-facing reads, and asking about an upcoming round is a natural thing for a client to do. It
gets a confident answer that is really just live state. A dilution before the boundary then
silently invalidates the client's index range. Please make the two consistent, or say that
`redistributionBatchAt` returns live state for future blocks and must not be used that way.

**2. `previousDepth != 0` overloads depth zero as "slot unwritten".** Safe today only because
`depth > bucketDepth >= 1` is enforced in `createBatch`/`copyBatch` — a dependency the SWIP
doesn't state. The failure mode is fail-closed (falls through to the smaller `olderDepth`, so
valid stamps get rejected rather than invalid ones accepted), but it is silent. Either state
the invariant, or use an explicit flag: `DepthHistory` is 18 bytes in a 32-byte slot, so two
bools cost nothing.

0xCardiE:
the SWIP gives a purpose for the six-round minimum but never derives the six?

lat-murmeldjur:
> Appendix B is a diff against a Phase 5 branch that doesn't exist or I am not aware of where this is. Was expecting that this SWIP is based on current phase 4 branch + swip 51 code

It's not based on any of those, swip49 is only relevant when stamp sampling is introduced, which is phase5 / swip50.

> setPrice is not claim-only, so the one-update assumption doesn't hold

Yes, it can happen in non-typical cases that the admin might want to adjust price. That is not the usualy flow, I think the clean solution is to only allow price setting manually when redistribution is paused - otherwise it would introduce serious complication of logic for tracking these states. If the redistribution is paused when manual price correction happens, the following round only should start with an unambigous state (possibly by a delay after unpause). 

> What needs deciding is whether SWIP-049's new import minimum is supposed to apply to migration

Similarly an edge case - minimum should not apply to migration, and/or it does not matter since the minimum is quite low timewise anyway (will expire in a few hours even if migrated).

> The Tolerance section's arithmetic doesn't survive SWIP-050

Correct, arithmetic is under update for SWIP-050 scheduling

> Two small things in Appendix A item 6

> 1. The two new views disagree on future blocks. _priceStateAtSamplingStart reverts
FutureSamplingStartBlock; redistributionBatchAt has no such guard, so a future block takes
the batch.lastUpdatedBlockNumber < samplingStartBlock branch and returns the current depth
as though it were historical.

In general the appendix is a non-normative code suggestion to help make implementation easier, any of these involved-detail misalignments can be fixed at some point. With this specific concern, we are adding the guard in the new iteration of the SWIP - however this would be fine probably without the guard as well as the bee client would be using this view as "which indexes should i use for the current sampling period" after the sampling period started, and not ask about a future sampling period.

> 2. previousDepth != 0 overloads depth zero as "slot unwritten". Safe today only because
depth > bucketDepth >= 1 is enforced in createBatch/copyBatch — a dependency the SWIP
doesn't state. The failure mode is fail-closed (falls through to the smaller olderDepth, so
valid stamps get rejected rather than invalid ones accepted), but it is silent. Either state
the invariant, or use an explicit flag: DepthHistory is 18 bytes in a 32-byte slot, so two
bools cost nothing.

Good observation. This is safe today. Zero depth batches would have zero indexes so they are not expected to be allowed in the future either. Probably sufficient to document this as a comment next to the relevant parts of the code. 


> the SWIP gives a purpose for the six-round minimum but never derives the six?

Six is quite arbitrary. As stated previously: 

"Choosing 912 block of minimum validity accepted for the result of batch operations:
Could be decreased to little more than 456 blocks. Otherwise this makes sure that batch operations can not move a batch in or out of the scope of valid batches for a round that already began sampling."

0xCardiE:
> > Appendix B is a diff against a Phase 5 branch that doesn't exist or I am not aware of where this is. Was expecting that this SWIP is based on current phase 4 branch + swip 51 code
> 
> It's not based on any of those, swip49 is only relevant when stamp sampling is introduced, which is phase5 / swip50.
> 
> > setPrice is not claim-only, so the one-update assumption doesn't hold
> 
> Yes, it can happen in non-typical cases that the admin might want to adjust price. That is not the usualy flow, I think the clean solution is to only allow price setting manually when redistribution is paused - otherwise it would introduce serious complication of logic for tracking these states. If the redistribution is paused when manual price correction happens, the following round only should start with an unambigous state (possibly by a delay after unpause).
> 
> > What needs deciding is whether SWIP-049's new import minimum is supposed to apply to migration
> 
> Similarly an edge case - minimum should not apply to migration, and/or it does not matter since the minimum is quite low timewise anyway (will expire in a few hours even if migrated).
> 
> > The Tolerance section's arithmetic doesn't survive SWIP-050
> 
> Correct, arithmetic is under update for SWIP-050 scheduling
> 
> > Two small things in Appendix A item 6
> 
> > 1. The two new views disagree on future blocks. _priceStateAtSamplingStart reverts
> >    FutureSamplingStartBlock; redistributionBatchAt has no such guard, so a future block takes
> >    the batch.lastUpdatedBlockNumber < samplingStartBlock branch and returns the current depth
> >    as though it were historical.
> 
> In general the appendix is a non-normative code suggestion to help make implementation easier, any of these involved-detail misalignments can be fixed at some point. With this specific concern, we are adding the guard in the new iteration of the SWIP - however this would be fine probably without the guard as well as the bee client would be using this view as "which indexes should i use for the current sampling period" after the sampling period started, and not ask about a future sampling period.
> 
> > 2. previousDepth != 0 overloads depth zero as "slot unwritten". Safe today only because
> >    depth > bucketDepth >= 1 is enforced in createBatch/copyBatch — a dependency the SWIP
> >    doesn't state. The failure mode is fail-closed (falls through to the smaller olderDepth, so
> >    valid stamps get rejected rather than invalid ones accepted), but it is silent. Either state
> >    the invariant, or use an explicit flag: DepthHistory is 18 bytes in a 32-byte slot, so two
> >    bools cost nothing.
> 
> Good observation. This is safe today. Zero depth batches would have zero indexes so they are not expected to be allowed in the future either. Probably sufficient to document this as a comment next to the relevant parts of the code.
> 
> > the SWIP gives a purpose for the six-round minimum but never derives the six?
> 
> Six is quite arbitrary. As stated previously:
> 
> "Choosing 912 block of minimum validity accepted for the result of batch operations: Could be decreased to little more than 456 blocks. Otherwise this makes sure that batch operations can not move a batch in or out of the scope of valid batches for a round that already began sampling."
```


function unPause() public {
    if (!hasRole(DEFAULT_ADMIN_ROLE, msg.sender)) revert OnlyPauser();
    if (block.number < uint256(PostageContract.lastUpdatedBlock()) + ROUND_LENGTH) {
        revert PriceSettlingPeriod();
    }
    _unpause();
}


```

I've taken the pause-first approach and made it enforceable rather than procedural:

- `PostageStamp` stores `previousPriceUpdatedBlock` and reverts `PriceHistoryUnavailable` when
  the rollback interval isn't covered by `previousPrice`  so an unresolvable history fails
  loudly instead of returning a wrong threshold.
- `Redistribution.unPause()` reverts `PriceSettlingPeriod` until a full round has passed since
  `PostageStamp.lastUpdatedBlock`, which is the concrete form of "a delay after unpause": it
  guarantees a sampling-start boundary sits between the price change and the first playable
  round.

I left "price changes only while paused" as an operational procedure rather than enforcing it,
since that would need PriceOracle to read Redistribution's paused state  new cross-contract
wiring for an event that should approach never.

Does that match what you had in mind, and is it worth folding the settling period into the SWIP?

PR #100, opened 2026-07-06
SWIP 050 - Sequential Transformation Scheme 1

lat-murmeldjur:
Sequential Transformation Scheme 1, or `STS-1`, extends Swarm redistribution from chunk-only sampling to a sequenced chunk-and-stamp proof. It keeps the existing running weighted truth-selection process, but an entry receives selection weight only after its selected stamp witnesses and STS-1 chunk-binding proofs pass.

### DIFF
--- a/SWIPs/swip-050.md
+++ b/SWIPs/swip-050.md
@@ -382,7 +382,7 @@ mapping(bytes32 => uint256) internal pendingCompletionIndexPlusOne;
 
 ### A.2 Round-bound commitment preimages
 
-The deployed `commit` function already accepts the commit round. STS only changes what the hidden hash commits to. The explicit round inside the hidden preimage prevents a commitment made for one round from being opened in a later round.
+The deployed `commit` function already accepts the commit round. STS only changes what the hidden hash commits to. The explicit round inside each hidden preimage prevents a commitment made for one round from being opened in a later round. The stamp sample hash commitment does not recommit `chunkSampleHash` or `chunkTransformRoot`; those values were already fixed by the stage-one chunk sample hash commitment and are stored after that reveal succeeds.
 
 ```solidity
 function wrapChunkSampleHashCommit(
@@ -408,9 +408,6 @@ function wrapChunkSampleHashCommit(
 function wrapStampSampleHashCommit(
     uint64 commitRound,
     bytes32 overlay,
-    uint8 depth,
-    bytes32 chunkSampleHash,
-    bytes32 chunkTransformRoot,
     bytes32 stampSampleHash,
     bytes32 revealNonce
 ) public pure returns (bytes32) {
@@ -418,9 +415,6 @@ function wrapStampSampleHashCommit(
         abi.encodePacked(
             commitRound,
             overlay,
-            depth,
-            chunkSampleHash,
-            chunkTransformRoot,
             stampSampleHash,
             revealNonce
         )
@@ -553,7 +547,7 @@ function markStageOneCommitmentProven(bytes32 overlay) internal {
 
 ### A.5 Stamp sample hash commit and reveal
 
-The stamp sample size is fixed at 16, so the stamp-sample commitment hides only `stampSampleHash` and the nonce, together with the already revealed stage-one values. The caller does not supply a reveal-array index. The STS extension stores the reveal index when the deployed reveal path accepts the stage-one reveal, and later STS functions recover the active `Reveal` through the caller overlay.
+The stamp sample size is fixed at 16, so the stamp-sample commitment hides only `stampSampleHash` and the nonce, bound to the commit round and overlay. It does not re-commit `chunkSampleHash`, `chunkTransformRoot`, or the claimed depth; those belong to the stage-one chunk sample hash commit and reveal. The stage-two commit is linked to stage one by accepting it only after the caller has a valid stage-one reveal for the same round, and by using the stored stage-one reveal data during stamp sample hash reveal and proof submission. The caller does not supply a reveal-array index. The STS extension stores the reveal index when the deployed reveal path accepts the stage-one reveal, and later STS functions recover the active `Reveal` through the caller overlay.
 
 ```solidity
 function revealIndexForOverlay(bytes32 overlay, uint64 roundNumber)
@@ -598,8 +592,7 @@ function revealStampSampleHash(
     if (commitRound != currentRevealRound) revert NoChunkSampleHashReveal();
 
     bytes32 overlay = Stakes.overlayOfAddress(msg.sender);
-    uint256 revealIndex = revealIndexForOverlay(overlay, commitRound);
-    Reveal storage revealRecord = currentReveals[revealIndex];
+    revealIndexForOverlay(overlay, commitRound);
     StsRevealData storage data = stsRevealOfOverlay[overlay];
     StampSampleCommitment storage stampCommit = stampCommitOfOverlay[overlay];
 
@@ -609,9 +602,6 @@ function revealStampSampleHash(
     bytes32 expected = wrapStampSampleHashCommit(
         commitRound,
         overlay,
-        revealRecord.depth,
-        revealRecord.hash,
-        data.chunkTransformRoot,
         stampSampleHash,
         revealNonce
     );
--- a/SWIPs/swip-050.md
+++ b/SWIPs/swip-050.md
@@ -1,0 +1,1100 @@
+---
+SWIP: TBD
+title: Sequential Transformation Scheme 1 for Redistribution
+author: TBD
+status: Draft
+type: Standards Track
+category: Storage Incentives
+created: 2026-06-30
+requires: SWIP-049
+---
+
+# SWIP TBD - Sequential Transformation Scheme 1 for Redistribution
+
+## Abstract
+
+Sequential Transformation Scheme 1, or `STS-1`, extends Swarm redistribution from chunk-only sampling to a sequenced chunk-and-stamp proof. It keeps the deployed truth-teller draw shape, but an entry receives selection weight only after its selected stamp witnesses and STS-1 chunk-binding proofs pass.
+
+The selected agreement value is:
+
+```text
+(chunkSampleHash, stampSampleHash, claimedDepth)
+```
+
+`chunkSampleHash` remains the chunk-side Schelling value. It encourages nodes to follow the same chunk rewrite and sample-construction conventions, but STS-1 does not open random chunk-sample witnesses and does not use chunk-sample density as a weight signal. Stamp density is the reliable density proof introduced by this SWIP.
+
+For round `r`, `firstAnchor(r)` is the deployed round anchor consumed by the round. It selects the neighbourhood and transforms chunk addresses. A valid chunk sample hash reveal updates the deployed seed for `firstAnchor(r + 1)` and derives `stampAnchor(r)`, which orders transformed stamp addresses for the current round. A later valid stamp sample hash reveal creates `proofSeed(r)`, which only selects stamp-sample positions to prove.
+
+Each selected stamp witness proves two links. First, the transformed stamp value is valid for `stampAnchor(r)`, the batch identifier, and the full stamp index, and the stamp signature proves that the batch owner authorized that stamp for a specific `chunkAddress`. Second, the transformed-chunk same-data proof verifies that a first-anchor transformed chunk address is derived from the same opened BMT data as that `chunkAddress`. The participant then proves that transformed address as a leaf inside its own `chunkTransformRoot`, committed in the first phase. Since `stampAnchor(r)` and `proofSeed(r)` are unknown when `chunkTransformRoot` is fixed, the participant cannot wait to learn the useful stamp range and add only those chunks later.
+
+After the truth teller selects the Schelling point, the selected round payout is divided among all proof-validated entries that reported the same Schelling point, in proportion to proof-adjusted stake density. Every paid node must prove separately because `chunkTransformRoot` is participant-specific and not part of the Schelling point.
+
+STS-1 preserves the deployed non-reveal freeze invariant across the longer sequence. A stage-one commit is unfinished until it becomes proof-validated. If its own round claims, that claim freezes unfinished entries using the selected truth depth. If its own round never claims, the unfinished entry is carried forward and finalized by the first later successful claim using the same duration rule.
+
+## Motivation
+
+The deployed mechanism already uses a transformed chunk-address sample as a Schelling value. STS-1 keeps that coordination role, but adds stamp sampling so depth overreporting depends on purchased stamp capacity. A transformed stamp address is computed from `stampAnchor(r)`, the batch identifier, and the full stamp index; a small witness set can verify batch existence, historical index range, signature, bucket alignment, balance, and liveness.
+
+Stamp metadata alone is not enough. A node could keep valid stamps while missing the corresponding chunks. STS-1 therefore accepts a selected stamp only when the transformed stamp value is valid for the stamp slot, the batch-owner signature is valid for the stamped `chunkAddress`, the transformed-chunk same-data proof verifies that a first-anchor transformed chunk address is derived from the same opened BMT data as that `chunkAddress`, and that transformed address was already included in the first-phase `chunkTransformRoot`. At the time that root is committed, the later stamp ordering and proof positions are unknown, so the node must already have the chunk to preserve the option of using the stamp later.
+
+The proposal addresses these issues:
+
+1. **Chunk/stamp separation.** A selected stamp counts only when its chunk was already part of the first-phase content-derived commitment.
+2. **Unproven selection.** A reveal has no truth-selection weight and no payout share until its stamp witnesses and binding proofs pass.
+3. **Minimum-only density.** A denser valid stamp sample receives a continuous benefit instead of a stepwise all-or-nothing benefit.
+4. **Batch utilization.** Lower within-bucket stamp indexes receive a continuous benefit because ordinary uploads tend to fill lower indexes before higher ones.
+
+The existing participation rule remains load-bearing: after a node commits, it must finish the required reveal path or be frozen by claim. This is the deployed non-reveal rule applied to the longer STS sequence, with carry-over only so that a round without a successful claim does not let unfinished STS commitments disappear.
+
+## Non-goals
+
+STS-1 does not change the total size or source of the redistribution pot. It changes how the selected round pot is distributed after the Schelling point is selected.
+
+STS-1 does not prove chunk-sample density. `chunkSampleHash` is a Schelling-point component, not a density witness path.
+
+STS-1 does not prove independent physical replicas. It proves that selected stamps refer to chunks in the participant's own earlier content-derived commitment.
+
+STS-1 relies on SWIP-049 for fixed historical postage scope. Node clients and contracts must evaluate the same batch and stamp-index set for the target round.
+
+## Definitions
+
+| Term | Meaning |
+|---|---|
+| `firstAnchor(r)` | The deployed round anchor consumed by round `r`; it selects the neighbourhood and transforms chunk addresses. It is produced by the existing seed, `currentRoundAnchor`, `currentSeed`, and skipped-round logic. |
+| `stampAnchor(r)` | STS-1 domain-separated anchor created from the updated deployed seed after the first valid chunk sample hash reveal in round `r`; it transforms stamp addresses for round `r`. |
+| `proofSeed(r)` | Randomness created by a valid stamp sample hash reveal; it selects stamp-sample positions to prove. |
+| `selectionSeed(r)` | Randomness used by the deployed-style truth-teller draw after proof submission. |
+| `chunkSampleHash` | Chunk-side Schelling value. STS-1 does not open random chunk-sample witnesses. |
+| `stampSampleHash` | Root or hash of the ordered transformed stamp-address sample. Selected stamp witnesses are opened against it. |
+| `chunkTransformRoot` | Participant-specific root over the claimed complete list of first-anchor transformed chunk addresses. |
+| `Schelling point` | `(chunkSampleHash, stampSampleHash, claimedDepth)`. |
+| `proof-validated entry` | Entry whose selected stamp witnesses and STS-1 binding proofs passed. |
+| `truthy proof-validated entry` | Proof-validated entry matching the selected Schelling point. |
+| `effectiveStakeDensity` | Deployed base stake density multiplied by the STS-1 stamp-density and utilization coefficients. |
+| `postageScopeStartBlock(r)` | SWIP-049 historical boundary for batch existence, depth, balance, and index range. |
+
+## Anchor and seed flow
+
+The anchor ownership rule is:
+
+```text
+round r - 1 reveal/skip path        -> deployed seed yields firstAnchor(r)
+round r                              -> consumes firstAnchor(r)
+round r chunk sample hash reveal     -> updates deployed seed for firstAnchor(r + 1) and derives stampAnchor(r)
+round r stamp sample hash reveal     -> creates proofSeed(r) and selectionSeed(r)
+```
+
+STS-1 does not add a second first-anchor storage system. `firstAnchor(r)` is the value returned by the deployed anchor path for round `r`. If round `r - 1` has no valid reveal, the deployed skipped-round path derives the next anchor by hashing the inherited seed once:
+
+```text
+firstAnchor(r) = H(firstAnchor(r - 1))
+```
+
+After `n` consecutive skipped rounds:
+
+```text
+firstAnchor(r + n) = H^n(firstAnchor(r))
+```
+
+A skipped round has no `stampAnchor`, no valid stamp sample, no `proofSeed`, and no STS-1 claim. The deterministic hash continuation is the deployed round-anchor continuation; STS-1 only consumes the resulting anchor.
+
+`stampAnchor(r)` is derived from the updated deployed seed produced by the first valid chunk sample hash reveal in round `r`. It is unknown during first-stage commitment, so the participant cannot choose `chunkTransformRoot` after seeing the stamp ordering.
+
+`proofSeed(r)` is created by the first valid stamp sample hash reveal in round `r`. It is unknown during stamp sample commitment, so a participant cannot precompute which stamp-sample positions will be opened.
+
+## Round structure and sampling windows
+
+An STS-1 round is 152 blocks. For a round starting at block `B`:
+
+| Blocks | Length | Phase |
+|---:|---:|---|
+| `B` to `B + 37` | 38 | chunk sample hash commit |
+| `B + 38` to `B + 56` | 19 | chunk sample hash reveal |
+| `B + 57` to `B + 94` | 38 | stamp sample hash commit |
+| `B + 95` to `B + 113` | 19 | stamp sample hash reveal |
+| `B + 114` to `B + 132` | 19 | proof submission |
+| `B + 133` to `B + 151` | 19 | truth selection, proportional payout settlement, and claim |
+
+The actual chunk-sampling time depends on when the previous round created `firstAnchor(r)`. The chunk commit deadline is `B + 38`.
+
+| Previous-round outcome | Chunk-sampling window |
+|---|---:|
+| Valid chunk sample hash reveal at `B - 114` | maximum: 152 blocks |
+| Valid chunk sample hash reveal at `B - 96` | minimum: 134 blocks |
+| Previous round skipped | maximum: 152 blocks |
+
+The actual stamp-sampling time depends on when the current round created `stampAnchor(r)`. The stamp commit deadline is `B + 95`.
+
+| Current-round outcome | Stamp-sampling window |
+|---|---:|
+| Valid chunk sample hash reveal at `B + 38` | maximum: 57 blocks |
+| Valid chunk sample hash reveal at `B + 56` | minimum: 39 blocks |
+| No valid chunk sample hash reveal | no valid stamp sample |
+
+SWIP-049 fixes the postage scope at `B - 114`, the earliest scheduled block where `firstAnchor(r)` may be fixed. This boundary is an eligibility boundary, not a promise that every node has exactly 152 blocks after learning the actual anchor.
+
+## Commit, hash-reveal, and proof sequence
+
+During chunk sample hash commit, the participant submits an obfuscated commitment for a specific commit round. The hidden preimage contains that round number, `claimedDepth`, `chunkSampleHash`, `chunkTransformRoot`, and a nonce. The round binding prevents a commitment prepared for one round from being opened in another.
+
+During chunk sample hash reveal, the participant opens that preimage with the same commit round. The contract checks the commitment and neighbourhood proximity against the deployed current reveal-round anchor. The first valid chunk sample hash reveal updates the deployed seed for `firstAnchor(r + 1)` and derives `stampAnchor(r)`. Without a valid chunk sample hash reveal, the round has no stamp proof path and the next first anchor comes from the deployed skipped-round hash continuation.
+
+During stamp sample hash commit, the participant commits to `stampSampleHash` and a nonce, again bound to the same round. The stamp sample size is fixed at 16 and is not participant-supplied data. The commit is accepted only after that participant has a valid chunk sample hash reveal for the same round.
+
+During stamp sample hash reveal, the participant opens the committed stamp sample hash. The first valid stamp sample hash reveal opens `proofSeed(r)` and `selectionSeed(r)`. There is no seed-only transaction that opens proof randomness while stamp sample hash reveals are still possible.
+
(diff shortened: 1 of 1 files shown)


significance:
I think there is an opportunity to simplify the protocol by separating **density** from **pre-knowledge**, as they are distinct security properties.

The density signal only needs to come from postage stamps. Since every valid stamp is cryptographically bound to exactly one chunk,

$$
\forall s \in S,\qquad s \mapsto c,
$$

a valid stamp proof already implies the existence of its corresponding chunk. Therefore, if the required stamp density is satisfied,

$$
|S_{\mathrm{valid}}| \ge D,
$$

then the existence of at least \(D\) qualifying chunks follows immediately. A separate chunk-density metric provides no additional information and is therefore redundant.

The chunk commitment serves a different purpose entirely. Its role is not to measure density, but to prove **prior possession** of the reserve. Let

$$
R_c = \mathrm{MerkleRoot}\left(\{T(c_i)\}\right),
$$

where \(T\) is a deterministic transformation (e.g. repeated Keccak hashing). For each sampled stamp \(s_i\), the participant proves

$$
s_i \rightarrow c_i,
\qquad
T(c_i) \in R_c.
$$

This establishes that the challenged chunk belonged to the participant's committed reserve before the challenge was known, preventing adaptive fetching after witness selection.

The protocol therefore decomposes cleanly into two independent proofs:

$$
\boxed{\mathrm{Stamp\ Proofs}\Longrightarrow\mathrm{Density}}
$$

$$
\boxed{\mathrm{Chunk\ Commitment}\Longrightarrow\mathrm{PreKnowledge}}
$$

The chunk commitment is therefore a **participant-specific commitment**, not a density signal or consensus object.

As a consequence, the second source of randomness and the additional commit/reveal phase appear unnecessary. A single commitment followed by a single unpredictable challenge is sufficient:

$$
\boxed{
\mathrm{Commit}
\longrightarrow
\mathrm{Random\ Challenge}
\longrightarrow
\mathrm{Prove}
}
$$

rather than

$$
\mathrm{Commit}
\longrightarrow
\mathrm{Random}
\longrightarrow
\mathrm{Commit}
\longrightarrow
\mathrm{Random}
\longrightarrow
\mathrm{Prove}.
$$

This preserves the anti-adaptive-storage property while significantly simplifying the protocol. It also cleanly separates the economic weighting mechanism (derived entirely from stamps) from the cryptographic proof of prior reserve possession (derived entirely from the participant's chunk commitment).

significance:
btw I also think that there may be better way to approach sampling in general - we need to not prove every segment and also could perhaps find a more efficient means of doing so

i also note the current cardinality of the sample is unworkably low and needs to be addressed and also that the consensus should be arrived at by an approach utilising the methodology outlined in my RS Fuzzy Sampling paper

0xCardiE:
Is the phase structure settled?
@significance raised that density comes from stamps alone, that the chunk commitment only proves prior possession, and that the second commit/reveal round is therefore unnecessary. That hasn't been answered.

We implemented the six phase version and it works end to end, but if it collapses to Commit → Random → Prove a fair bit of the plumbing moves. Worth settling before people build against it.

0xCardiE:
**sampleMaxValueForDepth isn't defined, and the depth argument looks wrong**
The function isn't defined anywhere in the SWIP, but the bigger issue is that it takes depth at all.

Today estimateSize compares the sample against a fixed sampleMaxValue. That fixed ceiling is what makes claiming a larger depth cost something. A node's reserve is capacity bounded so its 16th transformed value doesn't really move with depth, while stakeDensity grows as stake * 2^(depth - height).

If the ceiling scales with depth then overreporting by k bits gains 2^k on base density, L gains 2^k as well, so L/x is unchanged and the density coefficient doesn't react. Overreporting becomes free.

The Coefficient safety argument only works through underreporting. Is there an analysis for the other direction? We made the ceiling depth independent.

0xCardiE:
**Both coefficients are capped at 2^32, not 2x**
MAX_COEFFICIENT_Q64 = (1 << 32) * Q64. With bucketDepth 16 and a depth 32 batch there are 65536 slots per bucket, so three witnesses at index 0 give an average index ratio of 1/65536, a benefit ratio of 32768 and a 32x coefficient. Both coefficients multiply.

The tables stop at 2x and the text says worse utilization isn't penalised below 1x, which reads like the useful range is 1x to 2x.

As written, the cheapest way to multiply your weight by 32 is to buy a deep batch, stamp only index 0 of each bucket and never fill it. That rewards wasting capacity, which is the opposite of the stated motivation. It also happens to be exactly what a self stamper produces, since they choose their own indexes.

We capped both at 2x. Was 2^32 deliberate?

0xCardiE:
**bucketDepth < claimedDepth rejects every batch we have**
bucketDepth is fixed at creation and is 16 for essentially every mainnet batch. claimedDepth is already above 16 and goes up as the network grows. So this check rejects every real batch at any realistic depth and no proof could ever pass.

The neighbourhood binding it seems to be reaching for is already there twice: getPostageBucket(index) == addressToBucket(chunkAddress, bucketDepth) binds the stamp to the chunk, and inProximity(chunkAddress, firstAnchor, claimedDepth) binds the chunk to the neighbourhood.

We dropped it. Was it guarding something else?

0xCardiE:
**pendingCompletion is unbounded and can stop claim working permanently**
finalizeIncompleteCommitments walks the whole array in one claim transaction. Entries are added on every stage one commit and removed only by a successful proof or a successful claim.

STS-1 has five stages to clear, so any round where nothing claims carries its entire commit set forward. After enough of those the first successful claim runs out of gas, which means no claim can succeed, which means the array never drains. It doesn't need an attacker, a quiet stretch on a small network gets there.

SWIP-51 Option B already caps a round at MAX_COMMITS, and its finalize freezes non-finishers when the next round's first commit lands. So an unfinished participant is always resolved within one round, from an array that's bounded by construction.

We dropped pendingCompletion and made that path treat "revealed but never proved" the same as "never revealed". Does that lose anything the carry-over list was for?

lat-murmeldjur:
> Is the phase structure settled? @significance raised that density comes from stamps alone, that the chunk commitment only proves prior possession, and that the second commit/reveal round is therefore unnecessary. That hasn't been answered.
> 
> We implemented the six phase version and it works end to end, but if it collapses to Commit → Random → Prove a fair bit of the plumbing moves. Worth settling before people build against it.

The phase structure is settled. A different phase structure would form a different proposal different SWIP etc.

> density comes from stamps alone

yes however without the binding to chunks this density would only represent holding stamps, not storing chunks. The stage 1  chunk stage 2 stamp commitment is properly motivated, the second commit reveal is absolutely necessary.

> Commit → Random → Prove

as stated above that would be a different proposal, perhaps @sig will create that, but for the purposes of this proposal the 6 stage setup will remain.

lat-murmeldjur:
> **sampleMaxValueForDepth isn't defined, and the depth argument looks wrong** The function isn't defined anywhere in the SWIP, but the bigger issue is that it takes depth at all.
> 
> Today estimateSize compares the sample against a fixed sampleMaxValue. That fixed ceiling is what makes claiming a larger depth cost something. A node's reserve is capacity bounded so its 16th transformed value doesn't really move with depth, while stakeDensity grows as stake * 2^(depth - height).
> 
> If the ceiling scales with depth then overreporting by k bits gains 2^k on base density, L gains 2^k as well, so L/x is unchanged and the density coefficient doesn't react. Overreporting becomes free.
> 
> The Coefficient safety argument only works through underreporting. Is there an analysis for the other direction? We made the ceiling depth independent.

Yes. The big picture is, the appendix code suggestion is not reliable in the sense that some mistakes have not been spotted by me in the first generated version, this one of them.  (in any case the text part of  the proposal would be the more reliable)

In this case, the sampleMaxValue would remain unchanged, sampleMaxValueForDepth is in fact not necessary. estimateSize etc can be basically unchanged.

Okay so with coefficient safety, overreporting by +1 basically halves the usable entries, so overreporting is mostly only inhibited by the sampleMaxValue - it becomes less probable that overreporting still allows to sample low enough values if the number of usable entries is less than a minimal reserve.

lat-murmeldjur:
> **Both coefficients are capped at 2^32, not 2x** MAX_COEFFICIENT_Q64 = (1 << 32) * Q64. With bucketDepth 16 and a depth 32 batch there are 65536 slots per bucket, so three witnesses at index 0 give an average index ratio of 1/65536, a benefit ratio of 32768 and a 32x coefficient. Both coefficients multiply.
> 
> The tables stop at 2x and the text says worse utilization isn't penalised below 1x, which reads like the useful range is 1x to 2x.
> 
> As written, the cheapest way to multiply your weight by 32 is to buy a deep batch, stamp only index 0 of each bucket and never fill it. That rewards wasting capacity, which is the opposite of the stated motivation. It also happens to be exactly what a self stamper produces, since they choose their own indexes.
> 
> We capped both at 2x. Was 2^32 deliberate?

Coefficients should be capped at 32x. 
2x is not the limit the table is not showing every value just an illustration of a few of the early values up until 2x

lat-murmeldjur:
> **bucketDepth < claimedDepth rejects every batch we have** bucketDepth is fixed at creation and is 16 for essentially every mainnet batch. claimedDepth is already above 16 and goes up as the network grows. So this check rejects every real batch at any realistic depth and no proof could ever pass.
> 
> The neighbourhood binding it seems to be reaching for is already there twice: getPostageBucket(index) == addressToBucket(chunkAddress, bucketDepth) binds the stamp to the chunk, and inProximity(chunkAddress, firstAnchor, claimedDepth) binds the chunk to the neighbourhood.
> 
> We dropped it. Was it guarding something else?

Obvious error, my apologies for missing this, that check is not necessary.

Matching stamp index to neighborhood logic should be mostly unchanged, the used stamp indexes should be valid in the neighborhood defined by the reported depth and anchor

lat-murmeldjur:
> **pendingCompletion is unbounded and can stop claim working permanently** finalizeIncompleteCommitments walks the whole array in one claim transaction. Entries are added on every stage one commit and removed only by a successful proof or a successful claim.
> 
> STS-1 has five stages to clear, so any round where nothing claims carries its entire commit set forward. After enough of those the first successful claim runs out of gas, which means no claim can succeed, which means the array never drains. It doesn't need an attacker, a quiet stretch on a small network gets there.
> 
> SWIP-51 Option B already caps a round at MAX_COMMITS, and its finalize freezes non-finishers when the next round's first commit lands. So an unfinished participant is always resolved within one round, from an array that's bounded by construction.
> 
> We dropped pendingCompletion and made that path treat "revealed but never proved" the same as "never revealed". Does that lose anything the carry-over list was for?

Another error, sorry once again. 
What really should handle this after an unfinalized round (that has some commits not removed by proof, and with no claim to clear up), is the first commit in the next round. So when the first upcoming commit does this cleanup it cant accumulate more than 1 rounds commits.

0xCardiE:
Trying to determine do we expect Redistribution to keep any funds, and the only reason Redistribution would need to hold tokens at all is the pull payout.

Was pull deliberate? claim is already a pull, someone has to send that transaction for anything
to happen. So the appendix adds a second collection step: someone calls claim, then each node
calls withdraw separately.

I would do it the other way in the implementation. PostageStamp gains
withdrawShares(beneficiaries, weights) and transfers directly. Redistribution decides the split,
PostageStamp moves the money, no funds ever sit in Redistribution.

The one thing pull buys is that a recipient reverting on receipt can't block the others. BZZ has
no transfer hooks or blocklist so that seemed acceptable, and the current contract already pushes.

Anything I'm missing?

0xCardiE:
You mentioned "Reported Depth based coefficient should be capped at 2^16". I can't find it in the SWIP, only stampDensityCoefficient and utilizationCoefficient.

A 2^16 cap next to two 32x caps is a much wider range than the current coefficient safety argument covers.

Also worth knowing: Redistribution doesn't fit under EIP-170 with STS-1 inlined, at any optimizer setting. I had to move the witness verification and the Q64.64 maths into a deployed library to get under the limit. A third coefficient would add to that.

0xCardiE:
Two things I wanted to check:

Is a maximum reported depth an intended constraint of this proposal, or is there a path past it
later? Worth stating either way so client and batch tooling know.

Also the check is against the full reported depth rather than depth minus height. A node using
height reports a depth larger than its neighbourhood responsibility, so it reaches the ceiling
earlier than a height-0 node covering the same neighbourhood?

0xCardiE:
STS-1 drops chunk sample witnesses, so nothing checks chunkSampleHash against anything. A node
that can pass the stamp proofs, which needs real batches and real chunks but says nothing about
the chunk sample, can put an arbitrary value there. If its weight wins the truth draw then every
honest node in the neighbourhood disagrees and gets frozen while it takes the pot. Today that
isn't possible, the winner has to open three chunk witnesses against winner.hash.

The Rationale says it stays because it coordinates the chunk side convention among honest nodes.
That's a benefit, but being freezable on a value nobody checks is a cost, and Security
considerations doesn't mention it.

Keeping one chunk witness would close it. So would dropping chunkSampleHash from the truth tuple
and letting stampSampleHash carry the Schelling point alone.

0xCardiE:
claim has no msg.sender check, same as before, but the payoff changed. The winner used to take
the whole pot so it had an obvious reason to pay the gas. Now it splits N ways, so the caller
pays full gas for about a quarter of it at redundancy 4, an eighth at 8, or nothing at all if
they aren't a beneficiary.

If nobody calls, currentClaimRound doesn't advance, nothing is paid and the pot rolls forward.
Nobody is frozen for it, but everyone who proved loses that round.

Probably fine with a healthy pot, but every client needs a policy or they all wait for each
other. Is it meant to be any matching participant, first prover, lowest overlay?

0xCardiE:
chunkTransformRoot is fixed in stage one, but the stamp indexes are only derived once the stamp
anchor exists. If you build the tree in discovery order the root changes between those two points
and every membership proof fails. Leaves have to be in ascending order. I hit this in
implementation and every client will.

One sentence covering the tree construction would do: leaves ascending, parent is keccak of the
sorted pair, odd node promoted.

Separately, the text says the root is over the "complete" list of the reserve but nothing checks
completeness. Fine in practice since you can't predict what's useful, but "complete" probably
shouldn't be normative when it isn't enforced.

0xCardiE:
The proof seed mixes block.prevrandao at the block of the first valid stamp reveal. Whoever
reveals first picks that block, so they can look at the resulting seed, see which positions would
be opened, and not send the transaction, retrying later in the nineteen block window.

The sample is already committed so this doesn't let anyone invent one. It does let a node with a
partly supported sample steer away from its weak positions, which is what that section says it
prevents.

Accumulating prevrandao across all stamp reveals, the way updateRandomness already does for the
round seed, would remove it.

lat-murmeldjur:
> Trying to determine do we expect Redistribution to keep any funds, and the only reason Redistribution would need to hold tokens at all is the pull payout.
> 
> Was pull deliberate? claim is already a pull, someone has to send that transaction for anything to happen. So the appendix adds a second collection step: someone calls claim, then each node calls withdraw separately.
> 
> I would do it the other way in the implementation. PostageStamp gains withdrawShares(beneficiaries, weights) and transfers directly. Redistribution decides the split, PostageStamp moves the money, no funds ever sit in Redistribution.
> 
> The one thing pull buys is that a recipient reverting on receipt can't block the others. BZZ has no transfer hooks or blocklist so that seemed acceptable, and the current contract already pushes.
> 
> Anything I'm missing?

Yes actually the separate collection step is unnecessary - with claim we would ideally send "withdraw" transaction to all participants that reported the selected truth - proportionate to stake size.

Indeed redistribution does not need to hold funds.

lat-murmeldjur:
> You mentioned "Reported Depth based coefficient should be capped at 2^16". I can't find it in the SWIP, only stampDensityCoefficient and utilizationCoefficient.
> 
> A 2^16 cap next to two 32x caps is a much wider range than the current coefficient safety argument covers.
> 
> Also worth knowing: Redistribution doesn't fit under EIP-170 with STS-1 inlined, at any optimizer setting. I had to move the witness verification and the Q64.64 maths into a deployed library to get under the limit. A third coefficient would add to that.

Reported depth based coefficient is basically unchanged compared to the currently deployed contracts. The reason for the 2^16 is because of the bucket depth - with current bucket depth 16 the maximum reported storage depth we can accept is also 16 (where stamp density can be strictly checked still). Therefore 2^"reported depth" can maximum be 2^16.

Contract size fitting - good to know. Like Q64.64 is one option, but I'm not sure if any other alternatives would help bringing that down.

lat-murmeldjur:
> Two things I wanted to check:
> 
> Is a maximum reported depth an intended constraint of this proposal, or is there a path past it later? Worth stating either way so client and batch tooling know.
> 
> Also the check is against the full reported depth rather than depth minus height. A node using height reports a depth larger than its neighbourhood responsibility, so it reaches the ceiling earlier than a height-0 node covering the same neighbourhood?

Yes, the maximum reported storage depth is 16 - because of the current network wide bucket depths make stamps unambigous until this depth.

Good question about height. For any purposes we are not interested whether a node is storing another neighborhoods content - it should report the storage depth of a single neighborhood, and the "height" should only allow it to participate in the sybling neighborhood with one less Proximity Order to the anchor than the storage depth would require.

So PO(overlay, anchor) >= reported depth - height

lat-murmeldjur:
> STS-1 drops chunk sample witnesses, so nothing checks chunkSampleHash against anything. A node that can pass the stamp proofs, which needs real batches and real chunks but says nothing about the chunk sample, can put an arbitrary value there. If its weight wins the truth draw then every honest node in the neighbourhood disagrees and gets frozen while it takes the pot. Today that isn't possible, the winner has to open three chunk witnesses against winner.hash.
> 
> The Rationale says it stays because it coordinates the chunk side convention among honest nodes. That's a benefit, but being freezable on a value nobody checks is a cost, and Security considerations doesn't mention it.
> 
> Keeping one chunk witness would close it. So would dropping chunkSampleHash from the truth tuple and letting stampSampleHash carry the Schelling point alone.

No we don't want or need chunk witness. The aim of having the chunk sample at all is only to motivate following stamp rewrites.
If a stamp was already used, and the uploader reuses it for a different chunk, a node that has the old version of the stamp plus old chunk would be able to produce the same stamp sample as other nodes. However, because of this chunk sample being part of the schelling point, it now risks that using the old version of a chunk instead of the newest makes it fall off the schelling point of the honest nodes. Intentionally putting a random hash there is also leading to falling off the schelling point, and that is fine, because the aim is that the only safe strategy is trying to stay on the schelling point, anything else increases the risk of being frozen by other honest nodes that are polling their stakes by staying on the same schelling point.

lat-murmeldjur:
> claim has no msg.sender check, same as before, but the payoff changed. The winner used to take the whole pot so it had an obvious reason to pay the gas. Now it splits N ways, so the caller pays full gas for about a quarter of it at redundancy 4, an eighth at 8, or nothing at all if they aren't a beneficiary.
> 
> If nobody calls, currentClaimRound doesn't advance, nothing is paid and the pot rolls forward. Nobody is frozen for it, but everyone who proved loses that round.
> 
> Probably fine with a healthy pot, but every client needs a policy or they all wait for each other. Is it meant to be any matching participant, first prover, lowest overlay?

Good question, this is kind of undesigned - who should call the claim from all of the beneficiaries. For now we can probably leave it as it is - since everybody who is winning is motivated to make sure this is done, we can introduce a "convention view function" that tells a bee node whether it should call the claim - but still not restricting to that one participant only. Convention can be anything that is cheap to calculate, last node who proved from the winners sounds like one option.

lat-murmeldjur:
> chunkTransformRoot is fixed in stage one, but the stamp indexes are only derived once the stamp anchor exists. If you build the tree in discovery order the root changes between those two points and every membership proof fails. Leaves have to be in ascending order. I hit this in implementation and every client will.
> 
> One sentence covering the tree construction would do: leaves ascending, parent is keccak of the sorted pair, odd node promoted.
> 
> Separately, the text says the root is over the "complete" list of the reserve but nothing checks completeness. Fine in practice since you can't predict what's useful, but "complete" probably shouldn't be normative when it isn't enforced.

The order of chunks behind chunkTransformRoot is not strictly requiring any ordering (and in fact sorting can create more work).
The bee node can create the chunkTransformRoot in a random-discovery-order of chunks in the reserve that have a valid usable stamp by swip49 standards. While this chunkTransformRoot is committed, the file itself can be kept in memory/disc until the round is over.

Later the stamp sampling (also using the same range of stamps valid and usable by swip49 standards) gets the 16 lowest transformed stamp addresses by randomness 2. Now to find the matching entries in the chunktransform file, one only needs to create their transformed chunk addresses of the corresponding chunks of these 16 stamps using the randomness 1 and find the resulting transformed chunk addresses in the file for which the roottransformedhash was committed - by just checking where that transformed address is in the file, iterating through the 32 byte segments.

Also this can be left for optimization by sorting, or by recording the list of stamps in the same order when doing the initial chunk transformation iteration, and finding their corresponding transformed chunk address in stage 2 based on the index of the stamp in this array.

lat-murmeldjur:
> > Separately, the text says the root is over the "complete" list of the reserve but nothing checks completeness. Fine in practice since you can't predict what's useful, but "complete" probably shouldn't be normative when it isn't enforced.

Yes, good insight. Completeness is motivated by higher coefficients coming from higher density samples, and the aforementioned not being able to predict what will become useful.

lat-murmeldjur:
> The proof seed mixes block.prevrandao at the block of the first valid stamp reveal. Whoever reveals first picks that block, so they can look at the resulting seed, see which positions would be opened, and not send the transaction, retrying later in the nineteen block window.
> 
> The sample is already committed so this doesn't let anyone invent one. It does let a node with a partly supported sample steer away from its weak positions, which is what that section says it prevents.
> 
> Accumulating prevrandao across all stamp reveals, the way updateRandomness already does for the round seed, would remove it.

The reason why the first reveal is the source of the randomness - because that can be short-circuited by anyone who reveals early, removing any more entropy.

However always using a new randomness on a new reveal would allow as many rerandomizations as many reveals can be done - so one could have a number of commits and only reveal until a benefitting seed is created and accept the freeze for the non revealing identities. This would allow much more manipulation.

0xCardiE:
> > STS-1 drops chunk sample witnesses, so nothing checks chunkSampleHash against anything. A node that can pass the stamp proofs, which needs real batches and real chunks but says nothing about the chunk sample, can put an arbitrary value there. If its weight wins the truth draw then every honest node in the neighbourhood disagrees and gets frozen while it takes the pot. Today that isn't possible, the winner has to open three chunk witnesses against winner.hash.
> > The Rationale says it stays because it coordinates the chunk side convention among honest nodes. That's a benefit, but being freezable on a value nobody checks is a cost, and Security considerations doesn't mention it.
> > Keeping one chunk witness would close it. So would dropping chunkSampleHash from the truth tuple and letting stampSampleHash carry the Schelling point alone.
> 
> No we don't want or need chunk witness. The aim of having the chunk sample at all is only to motivate following stamp rewrites. If a stamp was already used, and the uploader reuses it for a different chunk, a node that has the old version of the stamp plus old chunk would be able to produce the same stamp sample as other nodes. However, because of this chunk sample being part of the schelling point, it now risks that using the old version of a chunk instead of the newest makes it fall off the schelling point of the honest nodes. Intentionally putting a random hash there is also leading to falling off the schelling point, and that is fine, because the aim is that the only safe strategy is trying to stay on the schelling point, anything else increases the risk of being frozen by other honest nodes that are polling their stakes by staying on the same schelling point.

That makes sense, thanks.

One thing that might be worth a line in Security considerations. With chunk witnesses a majority
attacker still had to commit to a provable sample, so honest nodes with the same reserve agreed
with it and weren't frozen. Now a majority attacker can commit garbage and guarantee everyone
else falls off. It needs majority weight in the neighbourhood, which already enables other
griefing, so it's a narrowing of an existing assumption rather than a new attack. But the
assumption did change.

0xCardiE:
> > Two things I wanted to check:
> > Is a maximum reported depth an intended constraint of this proposal, or is there a path past it later? Worth stating either way so client and batch tooling know.
> > Also the check is against the full reported depth rather than depth minus height. A node using height reports a depth larger than its neighbourhood responsibility, so it reaches the ceiling earlier than a height-0 node covering the same neighbourhood?
> 
> Yes, the maximum reported storage depth is 16 - because of the current network wide bucket depths make stamps unambigous until this depth.
> 
> Good question about height. For any purposes we are not interested whether a node is storing another neighborhoods content - it should report the storage depth of a single neighborhood, and the "height" should only allow it to participate in the sybling neighborhood with one less Proximity Order to the anchor than the storage depth would require.
> 
> So PO(overlay, anchor) >= reported depth - height
> 
> Edit: the reported depth based coefficient is also effected by height, in fact it should remain as it is now, 2^(reported depth-height)

Worth stating it in the proposal body rather than only in this thread, since client and batch tooling both need to know
the ceiling exists and why.

lat-murmeldjur:
> > > STS-1 drops chunk sample witnesses, so nothing checks chunkSampleHash against anything. A node that can pass the stamp proofs, which needs real batches and real chunks but says nothing about the chunk sample, can put an arbitrary value there. If its weight wins the truth draw then every honest node in the neighbourhood disagrees and gets frozen while it takes the pot. Today that isn't possible, the winner has to open three chunk witnesses against winner.hash.
> > > The Rationale says it stays because it coordinates the chunk side convention among honest nodes. That's a benefit, but being freezable on a value nobody checks is a cost, and Security considerations doesn't mention it.
> > > Keeping one chunk witness would close it. So would dropping chunkSampleHash from the truth tuple and letting stampSampleHash carry the Schelling point alone.
> > 
> > 
> > No we don't want or need chunk witness. The aim of having the chunk sample at all is only to motivate following stamp rewrites. If a stamp was already used, and the uploader reuses it for a different chunk, a node that has the old version of the stamp plus old chunk would be able to produce the same stamp sample as other nodes. However, because of this chunk sample being part of the schelling point, it now risks that using the old version of a chunk instead of the newest makes it fall off the schelling point of the honest nodes. Intentionally putting a random hash there is also leading to falling off the schelling point, and that is fine, because the aim is that the only safe strategy is trying to stay on the schelling point, anything else increases the risk of being frozen by other honest nodes that are polling their stakes by staying on the same schelling point.
> 
> That makes sense, thanks.
> 
> One thing that might be worth a line in Security considerations. With chunk witnesses a majority attacker still had to commit to a provable sample, so honest nodes with the same reserve agreed with it and weren't frozen. Now a majority attacker can commit garbage and guarantee everyone else falls off. It needs majority weight in the neighbourhood, which already enables other griefing, so it's a narrowing of an existing assumption rather than a new attack. But the assumption did change.

Yes that's a valid observation. The reason why we don't want to keep chunk witnesses, is because they are expensive (more inclusion proofs) and even if we would require them, an attacker that really wants to step off the schelling point and disagree could still create a different chunk sample anyway with little difficulty. Swip39 combined with fixed stakes should make it more difficult to achieve the majority weight.

0xCardiE:
<p dir="ltr"><strong>The freeze penalty got much bigger in two directions at once.</strong> A node now sends five transactions per round instead of two, and missing any one of the last four earns the harshest freeze, the same one a complete no-show gets. 

Because the depth is pinned at 16, that freeze will be in years time. 

image

We need to limit depth to lets say 10 or less so then we have something like 

image


There is also a systemic case: if nobody opens a stamp sample in a round, the proof seed never exists, nobody can prove even if they want to, and the whole neighborhood is frozen at the maximum rate.

0xCardiE:
Nothing requires the sixteen entries to be stamps on sixteen different chunks, and an empty Merkle proof makes a single-leaf chunk root valid, so in the limit all sixteen can point at one chunk. The cost of buying enough eligible stamps probably makes that uneconomic, but the SWIP should not describe the sample as evidence of reserve size without saying so.

lat-murmeldjur:
> **The freeze penalty got much bigger in two directions at once.** A node now sends five transactions per round instead of two, and missing any one of the last four earns the harshest freeze, the same one a complete no-show gets.

Okay so just to clarify this was not intentional, it's an error, freeze duration should be just like any other freeze duration, based on the depth of selected truth

lat-murmeldjur:
> Nothing requires the sixteen entries to be stamps on sixteen different chunks, and an empty Merkle proof makes a single-leaf chunk root valid, so in the limit all sixteen can point at one chunk. The cost of buying enough eligible stamps probably makes that uneconomic, but the SWIP should not describe the sample as evidence of reserve size without saying so.

It's not required that the stamps hold different chunks for multiple reasons. One is that multiple uploaders with different stamps and uploading the same chunk is a legit usecase. 

Evidence of the reserve size is in this case stamp wise, e.g. that the proving node has enough non-identical stamps to fill at least 50% of the reserve in the reported depth - even if all of the stamps could be holding a single chunk.

0xCardiE:
On the price update path, we hit a problem and implemented a guard. Want to check the
reasoning with you before it goes in the SWIP.

The problem. You said manual price setting should only happen while redistribution is
paused. The delay-after-unpause we added only constrains unpausing, so it does nothing
if the operator never pauses. If someone calls PriceOracle.setPrice in the middle of a
live round, that is a second price update inside that round. Every submitStsProof in
the round then reverts, because reading the price is the first thing it does. Nobody
becomes proof validated, the round has no truth, and the whole commit set gets frozen.
So the cost of forgetting to pause is a lost round plus a mass freeze.

What we did. PostageStamp.setPrice now refuses an update that would leave a round in
progress unable to resolve its own sampling start. The reasoning:

  - the rollback walks back exactly one price period, so a round is resolvable only
    while the update BEFORE the most recent one is at or before that round's sampling
    start
  - applying a new update pushes the current lastUpdatedBlock into that position
  - so the check is just: is lastUpdatedBlock already early enough for every round
    that can still read it

Which rounds can still read it? A round reads the price only while its witnesses are
verified. submitStsProof is the only caller and claim never touches it. So the proof
phase ending at offset 133 is the last moment a round can see an update. Before that
offset, the round in progress is still reading and the exposed sampling start is its
own, one round back. At or after it, the exposed round is the next one.

That one boundary separates the cases:

  update at r*152+133 (the oracle's claim)   -> accepted
  update at r*152+50  (mid round)            -> refused
  second update in the same round            -> refused
  update at r*152+50 after a quiet round     -> accepted

So your cadence always passes, and a manual change becomes available again a bit over
one round after claims stop, which is exactly what pausing produces. There is a
priceUpdatePermitted() view so an operator can check instead of guessing.

Four things I would like you to check:

1. Is offset 133 the right boundary? We are assuming nothing reads the price history
   after the proof phase ends. If anything else will read it later, the boundary moves.

2. The guard hard-codes the shape "one update per round, in the claim phase". That is
   what the oracle does today. Is that the cadence you intend to keep? If the phase
   table moves again, or the price is ever updated outside a claim, this breaks.

3. When PostageStamp refuses, PriceOracle has already advanced its own price and only
   emits StampPriceUpdateFailed. So the refused manual change gets pushed anyway on the
   next claim, at a block where the guard accepts it. It lands a round late instead of
   being rejected. Should a failed push roll the oracle back, so a manual set is
   all-or-nothing?

4. The check is conservative in one direction. If a round already claimed, or got no
   commits at all, PostageStamp cannot tell, so a mid-round update is still refused and
   the admin just waits. We think that is fine. Shout if not.

Also: this needs the claim-phase offset, 133, inside PostageStamp. That is SWIP-050's
phase table being used by SWIP-049's accounting. Worth deciding which document owns it.

0xCardiE:
> > Nothing requires the sixteen entries to be stamps on sixteen different chunks, and an empty Merkle proof makes a single-leaf chunk root valid, so in the limit all sixteen can point at one chunk. The cost of buying enough eligible stamps probably makes that uneconomic, but the SWIP should not describe the sample as evidence of reserve size without saying so.
> 
> It's not required that the stamps hold different chunks for multiple reasons. One is that multiple uploaders with different stamps and uploading the same chunk is a legit usecase.
> 
> Evidence of the reserve size is in this case stamp wise, e.g. that the proving node has enough non-identical stamps to fill at least 50% of the reserve in the reported depth - even if all of the stamps could be holding a single chunk.

Is the reserve 2^22? If it is, the default becomes 2^239. If "in the reported depth" means a different count at each depth, name the depth this constant is calibrated at one number cannot be 50% at all of them.

0xCardiE:
> btw I also think that there may be better way to approach sampling in general - we need to not prove every segment and also could perhaps find a more efficient means of doing so
> 
> i also note the current cardinality of the sample is unworkably low and needs to be addressed and also that the consensus should be arrived at by an approach utilising the methodology outlined in my RS Fuzzy Sampling paper

@lat-murmeldjur 

The stamp proof opens three of the sixteen sample entries: position 15 always, plus two drawn from 0..14. The other thirteen are never checked.

Someone who can back position 15 but pads the rest is caught with probability 1 - C(15-f, 2) / C(15, 2). One padded entry gets through 87% of the time, three get through 63%, five get through 43%. What that buys is one fewer genuine low transformed value needed to clear the density ceiling. The cost of being caught is the ordinary depth-based freeze.

Is sixteen entries and three openings the detection curve you want, or should more positions be opened?

0xCardiE:
The stamp anchor is derived after updateRandomness() advances the round seed, so the first stage-one revealer picks that block and can re-roll the anchor, the same way the first stamp reveal re-rolls the proof seed. The chunk-side anchor is fixed before that update.

Deriving the stamp anchor from currentRevealRoundAnchor instead removes that. Should it?

PR #93, opened 2026-04-30
Add: pubsub

nugaon:
A real-time messaging feature for dApps: WebSocket clients publish and subscribe to topic streams through Bee nodes, which act as the transport layer by leveraging their existing libp2p connections and bandwidth incentive system.

PR #103, opened 2026-08-02
SWIP-059: MEX — anchored message exchange with a previously unknown node

zelig:
## Summary

MEX (Message EXchange): a two-phase rendezvous-and-exchange pattern by which any node capable of push-sync and retrieval — including light, potentially non-reachable nodes — carries out a private, mutually authenticated message exchange with a **previously unknown** full node in the neighbourhood of a pre-agreed anchor. Built entirely from existing primitives: MIC/MOC (SWIP-42, #80), single-owner chunks, push-sync receipts, retrieval.

Extracted from the PubSub SWIP (#93), where it was worked out as Milestone 3, and generalised so it can be referenced independently — by PubSub as a dependency and by MIC/MOC as a use case — keeping the cryptographic detail out of the PubSub spec.

Also specifies:
- a **leakless known-responder variant** (PSS trojan first contact, two legs instead of three, reveals not even the fact of the exchange); the anchored and known-responder forms compose, so the anchor-side metadata leak is paid at most once (0–1×) per responder relationship
- security considerations incl. at-least-once delivery semantics ⇒ mandatory per-chunk-address session dedup (cf. [bee#5486 comment](https://github.com/ethersphere/bee/pull/5486#issuecomment-5150488841))

## Follow-ups
- [ ] trim Milestone 3 from #93 and reference this SWIP as a dependency instead
- [ ] SWIP-42 (#80) to list MEX as a use case

### DIFF
--- a/SWIPs/swip-59.md
+++ b/SWIPs/swip-59.md
@@ -1,5 +1,5 @@
 ---
-SWIP: 58
+SWIP: 59
 title: MEX — anchored message exchange with a previously unknown node
 author: Viktor Trón (@zelig), Viktor Tóth (@nugaon)
 discussions-to: https://discord.gg/Q6BvSkCv
--- a/SWIPs/swip-58.md
+++ b/SWIPs/swip-58.md
@@ -67,7 +67,7 @@ primitive with well-defined parameters, instead of near-copies with drifting det
 - **Service ID** `SID` — a well-known 32-byte SOC identifier that scopes the exchange to a
   service, e.g. `SID = keccak256("PUBSUB-REQUEST")`. Each MEX-based protocol MUST define
   its own `SID`.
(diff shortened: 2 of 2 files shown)


SWIPs/swip-59.md
```diff
@@ -0,0 +1,299 @@
...
+category: Networking
+created: 2026-07-31
+---
+
+<!-- Extracted and generalised from the PubSub SWIP (ethersphere/SWIPs PR #93), Milestone 3. -->
+
+## Simple Summary
+
+Any node capable of using the push-sync and retrieval protocols — including light nodes that
+are potentially non-reachable (NAT'd, sync-less) — can carry out a private, mutually
```
nugaon: I think here SWIP60 could be referenced that they are capable of pull syncing as well

PR #104, opened 2026-08-03
SWIP-060: BPS singlehop — brokered broadcast pub/sub, base protocol

zelig:
Base SWIP of the **Broadcast Pub/Sub (BPS)** family — the decomposition of the monolithic PubSub SWIP (#93) into work-package-sized SWIPs.

**What it specifies**: the smallest complete BPS protocol — one broker per topic, direct long-lived p2p streams, an explicit per-topic connection cap, SOC-only messages verified end-to-end. A cohort is fully described by a `CohortSpec` of genesis parameters; modes are parameter combinations, not an enum. Companion wire spec: [`assets/swip-60/bps.proto`](https://github.com/ethersphere/SWIPs/blob/swip-60-bps-singlehop/SWIPs/assets/swip-60/bps.proto) (singlehop concrete; multihop control frames reserved).

**Deliberately out of scope** (own SWIPs): multihop relaying/referral, reorganisation policies (SWATCH, SPORE), bandwidth incentives, broker discovery ([SWIP-59 MEX, #103](https://github.com/ethersphere/SWIPs/pull/103)), history delivery, implicit-publisher event sourcing.

**Relation to #93**: this SWIP absorbs its Milestone 1 plus the mode system (reframed as genesis parameters); Milestone 3 was already extracted as SWIP-59 (#103). Implementation groundwork: bee [#5435](https://github.com/ethersphere/bee/pull/5435), bee-js [#1151](https://github.com/ethersphere/bee-js/pull/1151).

### DIFF
--- a/SWIPs/assets/swip-60/bps.proto
+++ b/SWIPs/assets/swip-60/bps.proto
@@ -1,12 +1,21 @@
 // Broadcast Pub/Sub (BPS) — protocol messages and types.
 // Spec: SWIP-60 (../../swip-60.md).
 //
-// Deliberately incomplete as of 2026-08-02: the singlehop (depth = 1) subset is
-// concrete; multihop control-plane messages are named but reserved. The existing
-// implementation (bee PR #5435) uses hand-rolled byte framing with the same
-// semantics; this file is the normative description of the message structure,
-// and — bee protocols being protobuf-over-libp2p elsewhere — the candidate
-// replacement framing.
+// Revision 2 (2026-08-05), after review on PR #104: Connect split into
+// Open/Subscribe (subscribers carry no cohort metadata), broker capacity
+// removed from CohortSpec (it is broker-side policy, not a cohort parameter),
+// Ping dropped (liveness/RTT are transport concerns), and every frame carries
+// the full SOC (no handshake/data split). Field numbers renumbered — the
+// draft has no deployed compatibility surface.
+//
+// Enum zero values (*_UNSPECIFIED): proto3 requires a zero value; it is
+// deliberately NOT a legitimate wire value. It exists so that an unset field
+// is detectable and no implementation can silently rely on a default.
+// Receivers MUST reject messages carrying it.
+//
+// The singlehop (depth = 1) subset is concrete; multihop control-plane
+// messages are reserved. Implementation groundwork: bee PR #5435
+// (hand-rolled byte framing with the same semantics).
 
 syntax = "proto3";
 package bps;
@@ -17,50 +26,59 @@ option go_package = "github.com/ethersphere/bee/v2/pkg/bps/pb";
 // Cohort genesis — the primitive decisions whose combinations are the "modes"
 // ---------------------------------------------------------------------------
 
-// What the topic binds to (see epic: "What does the topic bind to?").
+// What the topic binds to (see SWIP-60: binding semantics).
 enum TopicBinding {
-  TOPIC_BINDING_UNSPECIFIED = 0;
+  TOPIC_BINDING_UNSPECIFIED = 0; // invalid on the wire (see header note)
   ANCHOR     = 1; // topic = full SOC/GSOC address; dedup on the wrapped CAC
   SOC_ID     = 2; // topic = SOC id; any owner with PO(addr, anchor) >= po_min
   OWNER      = 3; // topic = SOC owner; any id with PO(addr, anchor) >= po_min (MIC)
   FEED_TOPIC = 4; // id = keccak256(topic ‖ index); graffiti MIC / feed streams
 }
 
-// Who may author (see epic: genesis dimensions).
+// Who may author.
 enum PublisherRegime {
-  PUBLISHER_REGIME_UNSPECIFIED = 0;
-  EXPLICIT_SINGLE = 1; // opener is admin and sole publisher (live streaming)
-  EXPLICIT_LIST   = 2; // admin dictates who the other publishers are
+  PUBLISHER_REGIME_UNSPECIFIED = 0; // invalid on the wire (see header note)
+  EXPLICIT_SINGLE = 1; // opener is the sole publisher (live streaming)
+  EXPLICIT_LIST   = 2; // set fixed at genesis: admin + publisher_list
+                       // (dynamic grants/revocations: later revision)
   IMPLICIT        = 3; // authorship implied by the topic binding (PO constraint)
   ALL             = 4; // every peer publishes (gossipsub-equivalent cohort)
 }
 
-// The (partial) decisions fixed the moment the first full node is contacted.
+// Fixed by the cohort's opener; immutable for the cohort's lifetime.
+// NOTE: broker capacity is NOT a cohort parameter — a cohort cannot dictate a
+// remote node's connection count. Each broker enforces its own per-topic
+// stream limit and answers FULL when it is exhausted.
 message CohortSpec {
-  bytes           topic      = 1; // 32 bytes, meaning per binding
-  TopicBinding    binding    = 2;
-  PublisherRegime publishers = 3;
-  bool            history    = 4; // deliver matching chunks from the local store
-  bytes           admin      = 5; // 20-byte eth address; set iff EXPLICIT_*
-  uint32          po_min     = 6; // proximity order for implicit bindings (default 16)
-  uint32          cap        = 7; // max direct streams the broker accepts for this topic (0 = broker default)
-  bool            closed     = 8; // no audience: subscribers restricted to the publisher list
+  bytes           topic          = 1; // 32 bytes, meaning per binding
+  TopicBinding    binding        = 2;
+  PublisherRegime publishers     = 3;
+  bool            history        = 4; // deliver matching chunks from the local store
+  bytes           admin          = 5; // 20-byte eth address; set iff EXPLICIT_*
+  repeated bytes  publisher_list = 6; // 20-byte eth addresses, excl. admin;
+                                      // set iff EXPLICIT_LIST
+  uint32          po_min         = 7; // proximity order for implicit bindings (default 16)
+  bool            closed         = 8; // no audience: subscribers restricted to the publishers
 }
 
 // ---------------------------------------------------------------------------
-// Stream establishment (client -> broker), stream name "pubsub/1.0.0"
+// Stream establishment, stream name "pubsub/1.0.0" — one stream per (peer, topic).
+// The first message on a fresh stream is Open (fixes a new cohort) or
+// Subscribe (joins an existing one); the broker answers with Ack.
 // ---------------------------------------------------------------------------
 
-enum Role {
-  ROLE_UNSPECIFIED = 0;
-  SUBSCRIBER = 1;
-  PUBLISHER  = 2; // implies direct connection to the broker (necessary, not sufficient)
-}
-
-message Connect {
+// Opener -> broker: the one peer that fixes the cohort.
+message Open {
   CohortSpec    cohort = 1;
-  Role          role   = 2;
-  PublisherAuth auth   = 3; // present iff role == PUBLISHER
+  PublisherAuth auth   = 2; // present iff the opener publishes (explicit regimes)
+}
+
+// Joiner -> broker: names the topic — nothing more. Subscribers carry no
+// cohort metadata; auth is present iff the joiner publishes (publishers
+// connect directly to the broker).
+message Subscribe {
+  bytes         topic = 1; // 32 bytes
+  PublisherAuth auth  = 2; // present iff publisher
 }
 
 message PublisherAuth {
@@ -68,11 +86,30 @@ message PublisherAuth {
   bytes id    = 2; // 32-byte SOC id, when the binding fixes it
 }
 
+// Broker -> peer, answering Open or Subscribe. The echoed CohortSpec lets a
+// subscriber verify every message end-to-end against the topic binding.
+message Ack {
+  Status     status = 1;
+  CohortSpec cohort = 2; // set iff status == OK
+}
+
+enum Status {
+  STATUS_UNSPECIFIED = 0; // invalid on the wire (see header note)
+  OK                 = 1;
+  FULL               = 2; // broker at its per-topic capacity;
+                          // a singlehop broker refuses — nothing else
+  UNKNOWN_TOPIC      = 3; // Subscribe for a topic the broker does not serve
+  REJECTED           = 4; // e.g. publisher not on the list, invalid auth,
+                          // non-publisher Subscribe on a closed cohort
+}
+
 // ---------------------------------------------------------------------------
 // Messages — SOC-only is a protocol feature
 // ---------------------------------------------------------------------------
 
-// A full single-owner chunk in transit.
+// A full single-owner chunk in transit. Every frame is self-contained: no
+// per-stream handshake state, and no format change if the stream model
+// evolves (e.g. topic-muxed streams later).
 message Soc {
   bytes id        = 1; // 32 bytes
   bytes owner     = 2; // 20 bytes (recoverable from signature; explicit for cheap filtering)
@@ -81,33 +118,20 @@ message Soc {
   bytes payload   = 5; // wrapped-CAC data, <= 4096 bytes
 }
 
-// Publisher -> broker. No type prefix needed: the stream's role was declared at Connect.
+// Publisher -> broker.
 message Publish {
   Soc soc = 1;
 }
 
-// Broker -> subscriber: exactly one of the following per frame.
+// Broker -> subscriber.
 message Broadcast {
   oneof frame {
-    Soc       handshake = 1; // first frame on a stream: full SOC identity
-    DataFrame data      = 2; // subsequent frames: signature ‖ span ‖ payload only
-    Ping      ping      = 3; // keepalive; parent measures RTT off the echo
+    Soc soc = 1;
+    // 2–15 reserved: multihop control plane (Beacon, Reparent, Expect,
+    // DcutrSignal, SwapProposal) — named to fix intent, not final.
   }
 }
 
-message DataFrame {
-  bytes signature = 1;
-  bytes span      = 2;
-  bytes payload   = 3;
-}
-
-message Ping {}
-
-// ---------------------------------------------------------------------------
-// Multihop control plane — RESERVED, named to fix intent (not final for AFM)
-// ---------------------------------------------------------------------------
-// message Beacon        {}  // child -> parent capacity/score summary (0xFE)
-// message Reparent      {}  // parent -> child: REPARENT{to, gateway?}   (0xFD)
-// message Expect        {}  // parent -> relay: EXPECT{children}         (0xFC)
-// message DcutrSignal   {}  // via circuit relay                          (0xFB)
-// message SwapProposal  {}  // promotion swap propose/ack                 (0xFA)
+// Keepalive / RTT: none at the BPS level. Liveness is the transport's job
+// (libp2p), and latency metrics for reorganisation policies (SWATCH) are
+// sourced there as well.
--- a/SWIPs/assets/swip-60/bps.proto
+++ b/SWIPs/assets/swip-60/bps.proto
@@ -30,8 +30,8 @@ option go_package = "github.com/ethersphere/bee/v2/pkg/bps/pb";
 enum TopicBinding {
   TOPIC_BINDING_UNSPECIFIED = 0; // invalid on the wire (see header note)
   ANCHOR     = 1; // topic = full SOC/GSOC address; dedup on the wrapped CAC
-  SOC_ID     = 2; // topic = SOC id; any owner with PO(addr, anchor) >= po_min
-  OWNER      = 3; // topic = SOC owner; any id with PO(addr, anchor) >= po_min (MIC)
+  SOC_ID     = 2; // topic = SOC id; any owner with PO(addr, anchor) >= PO_MIN
+  OWNER      = 3; // topic = SOC owner; any id with PO(addr, anchor) >= PO_MIN (MIC)
   FEED_TOPIC = 4; // id = keccak256(topic ‖ index); graffiti MIC / feed streams
 }
 
@@ -49,6 +49,10 @@ enum PublisherRegime {
 // NOTE: broker capacity is NOT a cohort parameter — a cohort cannot dictate a
 // remote node's connection count. Each broker enforces its own per-topic
 // stream limit and answers FULL when it is exhausted.
+// NOTE: the proximity constraint for implicit bindings is a protocol
+// constant, PO_MIN = 16 — not a cohort parameter (a proto3 unset uint32 is
+// indistinguishable from 0, which would silently disable the constraint;
+// and no use case varies it).
 message CohortSpec {
   bytes           topic          = 1; // 32 bytes, meaning per binding
   TopicBinding    binding        = 2;
@@ -57,7 +61,7 @@ message CohortSpec {
   bytes           admin          = 5; // 20-byte eth address; set iff EXPLICIT_*
   repeated bytes  publisher_list = 6; // 20-byte eth addresses, excl. admin;
                                       // set iff EXPLICIT_LIST
-  uint32          po_min         = 7; // proximity order for implicit bindings (default 16)
+  reserved 7;                         // was po_min — now protocol constant PO_MIN
   bool            closed         = 8; // no audience: subscribers restricted to the publishers
 }
 
--- a/SWIPs/assets/swip-60/bps.proto
+++ b/SWIPs/assets/swip-60/bps.proto
@@ -1,0 +1,113 @@
+// Broadcast Pub/Sub (BPS) — protocol messages and types.
+// Spec: SWIP-60 (../../swip-60.md).
+//
+// Deliberately incomplete as of 2026-08-02: the singlehop (depth = 1) subset is
+// concrete; multihop control-plane messages are named but reserved. The existing
+// implementation (bee PR #5435) uses hand-rolled byte framing with the same
+// semantics; this file is the normative description of the message structure,
+// and — bee protocols being protobuf-over-libp2p elsewhere — the candidate
+// replacement framing.
+
+syntax = "proto3";
+package bps;
+
+option go_package = "github.com/ethersphere/bee/v2/pkg/bps/pb";
+
+// ---------------------------------------------------------------------------
+// Cohort genesis — the primitive decisions whose combinations are the "modes"
+// ---------------------------------------------------------------------------
+
+// What the topic binds to (see epic: "What does the topic bind to?").
+enum TopicBinding {
+  TOPIC_BINDING_UNSPECIFIED = 0;
+  ANCHOR     = 1; // topic = full SOC/GSOC address; dedup on the wrapped CAC
+  SOC_ID     = 2; // topic = SOC id; any owner with PO(addr, anchor) >= po_min
+  OWNER      = 3; // topic = SOC owner; any id with PO(addr, anchor) >= po_min (MIC)
+  FEED_TOPIC = 4; // id = keccak256(topic ‖ index); graffiti MIC / feed streams
+}
+
+// Who may author (see epic: genesis dimensions).
+enum PublisherRegime {
+  PUBLISHER_REGIME_UNSPECIFIED = 0;
+  EXPLICIT_SINGLE = 1; // opener is admin and sole publisher (live streaming)
+  EXPLICIT_LIST   = 2; // admin dictates who the other publishers are
+  IMPLICIT        = 3; // authorship implied by the topic binding (PO constraint)
+  ALL             = 4; // every peer publishes (gossipsub-equivalent cohort)
+}
+
+// The (partial) decisions fixed the moment the first full node is contacted.
+message CohortSpec {
+  bytes           topic      = 1; // 32 bytes, meaning per binding
+  TopicBinding    binding    = 2;
+  PublisherRegime publishers = 3;
+  bool            history    = 4; // deliver matching chunks from the local store
+  bytes           admin      = 5; // 20-byte eth address; set iff EXPLICIT_*
+  uint32          po_min     = 6; // proximity order for implicit bindings (default 16)
+  uint32          cap        = 7; // max direct streams the broker accepts for this topic (0 = broker default)
+  bool            closed     = 8; // no audience: subscribers restricted to the publisher list
+}
+
+// ---------------------------------------------------------------------------
+// Stream establishment (client -> broker), stream name "pubsub/1.0.0"
+// ---------------------------------------------------------------------------
+
+enum Role {
+  ROLE_UNSPECIFIED = 0;
+  SUBSCRIBER = 1;
+  PUBLISHER  = 2; // implies direct connection to the broker (necessary, not sufficient)
+}
+
+message Connect {
+  CohortSpec    cohort = 1;
+  Role          role   = 2;
+  PublisherAuth auth   = 3; // present iff role == PUBLISHER
+}
+
+message PublisherAuth {
+  bytes owner = 1; // 20-byte eth address of the SOC owner key
+  bytes id    = 2; // 32-byte SOC id, when the binding fixes it
+}
+
+// ---------------------------------------------------------------------------
+// Messages — SOC-only is a protocol feature
+// ---------------------------------------------------------------------------
+
+// A full single-owner chunk in transit.
+message Soc {
+  bytes id        = 1; // 32 bytes
+  bytes owner     = 2; // 20 bytes (recoverable from signature; explicit for cheap filtering)
+  bytes signature = 3; // 65 bytes
+  bytes span      = 4; // 8 bytes LE
+  bytes payload   = 5; // wrapped-CAC data, <= 4096 bytes
+}
+
+// Publisher -> broker. No type prefix needed: the stream's role was declared at Connect.
+message Publish {
+  Soc soc = 1;
+}
+
+// Broker -> subscriber: exactly one of the following per frame.
+message Broadcast {
+  oneof frame {
+    Soc       handshake = 1; // first frame on a stream: full SOC identity
+    DataFrame data      = 2; // subsequent frames: signature ‖ span ‖ payload only
+    Ping      ping      = 3; // keepalive; parent measures RTT off the echo
+  }
+}
+
+message DataFrame {
+  bytes signature = 1;
+  bytes span      = 2;
+  bytes payload   = 3;
+}
+
+message Ping {}
+
+// ---------------------------------------------------------------------------
+// Multihop control plane — RESERVED, named to fix intent (not final for AFM)
+// ---------------------------------------------------------------------------
+// message Beacon        {}  // child -> parent capacity/score summary (0xFE)
+// message Reparent      {}  // parent -> child: REPARENT{to, gateway?}   (0xFD)
+// message Expect        {}  // parent -> relay: EXPECT{children}         (0xFC)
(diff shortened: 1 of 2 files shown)


SWIPs/assets/swip-60/bps.proto
```diff
@@ -0,0 +1,113 @@
...
+
+option go_package = "github.com/ethersphere/bee/v2/pkg/bps/pb";
+
+// ---------------------------------------------------------------------------
+// Cohort genesis — the primitive decisions whose combinations are the "modes"
+// ---------------------------------------------------------------------------
+
+// What the topic binds to (see epic: "What does the topic bind to?").
+enum TopicBinding {
+  TOPIC_BINDING_UNSPECIFIED = 0;
```
acud: what does this semantically mean? why is this a legitimate value that can be used?

SWIPs/assets/swip-60/bps.proto
```diff
@@ -0,0 +1,113 @@
...
+  ANCHOR     = 1; // topic = full SOC/GSOC address; dedup on the wrapped CAC
+  SOC_ID     = 2; // topic = SOC id; any owner with PO(addr, anchor) >= po_min
+  OWNER      = 3; // topic = SOC owner; any id with PO(addr, anchor) >= po_min (MIC)
+  FEED_TOPIC = 4; // id = keccak256(topic ‖ index); graffiti MIC / feed streams
+}
+
+// Who may author (see epic: genesis dimensions).
+enum PublisherRegime {
+  PUBLISHER_REGIME_UNSPECIFIED = 0;
```
acud: what does this semantically mean? why is this a legitimate value that can be used?

SWIPs/assets/swip-60/bps.proto
```diff
@@ -0,0 +1,113 @@
...
+  ALL             = 4; // every peer publishes (gossipsub-equivalent cohort)
+}
+
+// The (partial) decisions fixed the moment the first full node is contacted.
+message CohortSpec {
+  bytes           topic      = 1; // 32 bytes, meaning per binding
+  TopicBinding    binding    = 2;
+  PublisherRegime publishers = 3;
+  bool            history    = 4; // deliver matching chunks from the local store
+  bytes           admin      = 5; // 20-byte eth address; set iff EXPLICIT_*
```
acud: if `EXPLICIT_LIST` is this then a concatenated list of ethereum keys?

SWIPs/assets/swip-60/bps.proto
```diff
@@ -0,0 +1,113 @@
...
+  FEED_TOPIC = 4; // id = keccak256(topic ‖ index); graffiti MIC / feed streams
+}
+
+// Who may author (see epic: genesis dimensions).
+enum PublisherRegime {
+  PUBLISHER_REGIME_UNSPECIFIED = 0;
+  EXPLICIT_SINGLE = 1; // opener is admin and sole publisher (live streaming)
+  EXPLICIT_LIST   = 2; // admin dictates who the other publishers are
+  IMPLICIT        = 3; // authorship implied by the topic binding (PO constraint)
+  ALL             = 4; // every peer publishes (gossipsub-equivalent cohort)
```
acud: why do we need this?

SWIPs/assets/swip-60/bps.proto
```diff
@@ -0,0 +1,113 @@
...
+
+// The (partial) decisions fixed the moment the first full node is contacted.
+message CohortSpec {
+  bytes           topic      = 1; // 32 bytes, meaning per binding
+  TopicBinding    binding    = 2;
+  PublisherRegime publishers = 3;
+  bool            history    = 4; // deliver matching chunks from the local store
+  bytes           admin      = 5; // 20-byte eth address; set iff EXPLICIT_*
+  uint32          po_min     = 6; // proximity order for implicit bindings (default 16)
+  uint32          cap        = 7; // max direct streams the broker accepts for this topic (0 = broker default)
```
acud: why should a 3rd party be able to control the number of connections on a remote peer? what if the number is too large for the remote node to accept?

SWIPs/assets/swip-60/bps.proto
```diff
@@ -0,0 +1,113 @@
...
+  bool            closed     = 8; // no audience: subscribers restricted to the publisher list
+}
+
+// ---------------------------------------------------------------------------
+// Stream establishment (client -> broker), stream name "pubsub/1.0.0"
+// ---------------------------------------------------------------------------
+
+enum Role {
+  ROLE_UNSPECIFIED = 0;
```
acud: why is this a legitimate value?

SWIPs/assets/swip-60/bps.proto
```diff
@@ -0,0 +1,113 @@
...
+enum PublisherRegime {
+  PUBLISHER_REGIME_UNSPECIFIED = 0;
+  EXPLICIT_SINGLE = 1; // opener is admin and sole publisher (live streaming)
+  EXPLICIT_LIST   = 2; // admin dictates who the other publishers are
+  IMPLICIT        = 3; // authorship implied by the topic binding (PO constraint)
+  ALL             = 4; // every peer publishes (gossipsub-equivalent cohort)
+}
+
+// The (partial) decisions fixed the moment the first full node is contacted.
+message CohortSpec {
```
acud, line 70: the comment is misleading and partially incorrect. the seq diagram in the markdown file defines that actually both publisher _and_ subscriber use the same type of message to connect to a broker. i'm not sure what are my feelings around this. it seems to be too elaborate for both sides to use symmetrically - why does a subscriber need to provide the whole `CohortSpec` when connecting? i.e. why is it necessary to mention the `admin`, `history`, `publishers`, etc? seems like irrelevant information. i can't see why a subscriber should provide anything more than a topic and an identity, and that's it.

so it begs the question, why shouldn't a state channel opening become its own specific message? and then equally `Connect` turns into a `Subscribe`

SWIPs/assets/swip-60/bps.proto
```diff
@@ -0,0 +1,113 @@
...
+// Publisher -> broker. No type prefix needed: the stream's role was declared at Connect.
+message Publish {
+  Soc soc = 1;
+}
+
+// Broker -> subscriber: exactly one of the following per frame.
+message Broadcast {
+  oneof frame {
+    Soc       handshake = 1; // first frame on a stream: full SOC identity
+    DataFrame data      = 2; // subsequent frames: signature ‖ span ‖ payload only
```
acud: why split the same chunk to multiple messages? you're also overloading the protocol code to do the message sequencing/buffering/etc... seems really unnecessary. also if you assume just one stream per topic then essentially you're coercing the applications to manage multiple streams between the same two peers continuously - why not multiplex everything over the same stream?

SWIPs/assets/swip-60/bps.proto
```diff
@@ -0,0 +1,113 @@
...
+    Ping      ping      = 3; // keepalive; parent measures RTT off the echo
```
acud: who needs this information? we already measure rtt using other means. not sure why this message is needed

SWIPs/swip-60.md
```diff
@@ -0,0 +1,242 @@
...
+cohort's `CohortSpec` scopes every session; authentication is structural (SOC-signed
+against the topic binding), so brokers and relays forward without being trusted — an
+intermediate can withhold, never forge; and withholding is a liveness fault recoverable
+by re-pointing or relocating the topic. Multihop forwarding (bps-multihop) adds capacity
+without reintroducing flooding: every edge still pays upstream, every node still receives
+only its topic's stream.
+
+## Out of scope (deliberately)
+
+Multihop relaying and referral (bps-multihop), reorganisation policies (SWATCH, SPORE —
```
acud, line 648: i would also add to this: remove multi-publisher setup from this iteration. it can be added later and just adds more review surface to deal with at this moment.

SWIPs/assets/swip-60/bps.proto
```diff
@@ -0,0 +1,113 @@
...
+  SOC_ID     = 2; // topic = SOC id; any owner with PO(addr, anchor) >= po_min
+  OWNER      = 3; // topic = SOC owner; any id with PO(addr, anchor) >= po_min (MIC)
+  FEED_TOPIC = 4; // id = keccak256(topic ‖ index); graffiti MIC / feed streams
+}
+
+// Who may author (see epic: genesis dimensions).
+enum PublisherRegime {
+  PUBLISHER_REGIME_UNSPECIFIED = 0;
+  EXPLICIT_SINGLE = 1; // opener is admin and sole publisher (live streaming)
+  EXPLICIT_LIST   = 2; // admin dictates who the other publishers are
```
acud: i would get rid of this for a first iteration

SWIPs/assets/swip-60/bps.proto
```diff
@@ -0,0 +1,113 @@
...
+package bps;
+
+option go_package = "github.com/ethersphere/bee/v2/pkg/bps/pb";
+
+// ---------------------------------------------------------------------------
+// Cohort genesis — the primitive decisions whose combinations are the "modes"
+// ---------------------------------------------------------------------------
+
+// What the topic binds to (see epic: "What does the topic bind to?").
+enum TopicBinding {
```
acud, line 39: nit: i find this whole thing really confusing and not very approachable and i wonder if this even makes sense to do in a first iteration. "pubsub" is very dumb in this sense - it usually does not give you different topic semantics. here, a topic could have different semantics and input validation according to its "type" which makes for a much more complex API surfaces for users later on...

SWIPs/assets/swip-60/bps.proto
acud: while the single/multiple stream part is debatable, i'm not sure we need to skimp out on these few bytes that the chunk carries - it really doesn't save much, and then if we want to do single/multi stream management, we don't have to break the message format

SWIPs/assets/swip-60/bps.proto
zelig: - confusing, not very approachable, does not make sense, very dumb, no topic semantics, hmmm, thats a lot of negative things to asspciated to something that could  have different semantics according to its type which makes for a... complex API surfaces? hhwhhat?

SWIPs/assets/swip-60/bps.proto
zelig: Alright , but then you cannot get rid of ALL, otherwise you cannot have collab eediting..see below

SWIPs/assets/swip-60/bps.proto
zelig: either ALL or EXPLICIT list needed. HOnestly I do not find it very natural that you can edit a file ith 3 other random people :) you want to restrict, explicitly list those that do .

SWIPs/assets/swip-60/bps.proto
zelig: fair enough i dont think it should

SWIPs/assets/swip-60/bps.proto
zelig: fair, it is not

SWIPs/swip-60.md
zelig: funnily i think a single one only is more complex to implement since you need to authenticate.

SWIPs/assets/swip-60/bps.proto
zelig: it becomes clear by now that this is a pattern. you do not want devs to use a default zero value, ie., leave it out of the struct defined.

zelig:
Revision 2 pushed (25f6f08), addressing the review:

**Taken:**
- `Connect` split into **`Open`** (the opener fixes the `CohortSpec`) and **`Subscribe`** (topic only — subscribers carry no cohort metadata). The broker's `Ack` echoes the spec back so subscribers can verify every message end-to-end. The `Role` enum is gone with the split.
- **Capacity is out of `CohortSpec`** — agreed a cohort shouldn't dictate a remote node's connection count. It's broker-side policy now; `FULL` semantics unchanged. The jam-cohort seat bound comes from the genesis publisher list instead.
- **Handshake/data frame split dropped** — every frame carries the full SOC, so no sequencing state and no format break if the stream model changes later. Kept one stream per (peer, topic) for now, with the rationale spelled out in the wire section (per-cohort flow control/teardown, bee protocol idiom; mux migration stays format-compatible).
- **Ping dropped** — liveness/RTT are transport concerns.
- `*_UNSPECIFIED = 0`: documented in the proto header — proto3 requires a zero value; it is deliberately invalid on the wire so nothing can rely on a default.

**Specified (was a gap):** `EXPLICIT_LIST` is a `repeated publisher_list` fixed at genesis — not concatenated bytes. Dynamic grants/revocations are explicitly out of scope for this iteration, which trims the review surface without dropping multi-publisher.

**Kept, per the discussion above:** `TopicBinding` and the multi-publisher regimes (`EXPLICIT_LIST`/`ALL`) — the closed collab-editing cohort is a headline use case, and single-publisher-only wouldn't even be simpler, since publisher authentication is needed regardless.

SWIPs/swip-60.md
```diff
@@ -0,0 +1,256 @@
...
+no mode enum; **modes are combinations of these parameters**.
+
+| parameter | values | meaning |
+|---|---|---|
+| `topic` | 32 bytes | interpreted per `binding` |
+| `binding` | `ANCHOR` / `SOC_ID` / `OWNER` / `FEED_TOPIC` | what the topic binds to; fixes which SOCs qualify as messages and the dedup rule |
+| `publishers` | `EXPLICIT_SINGLE` / `EXPLICIT_LIST` / `IMPLICIT` / `ALL` | who may author |
+| `admin` + `publisher_list` | eth addresses | set iff explicit publishers; with `EXPLICIT_LIST` the full publisher set is **fixed at genesis** (dynamic grants/revocations are deferred to a later revision) |
+| `history` | bool | deliver matching chunks already in the local store (mechanism in bps-history; a singlehop broker MAY refuse) |
+| `po_min` | uint (default 16) | proximity constraint for implicit bindings: `PO(socAddr, anchor) ≥ po_min` |
```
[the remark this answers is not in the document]
zelig: po_min should be  a constant not. a param

SWIPs/assets/swip-60/bps.proto
acud: i meant the concept of [pubsub](https://en.wikipedia.org/wiki/Publish%E2%80%93subscribe_pattern) usually does not offer different semantics over the concept of a topic. i would appreciate you not hijacking my words and initial intention as this is really counter productive and aggressive. thanks

zelig: I quoted your words which indeed were unnecessarily agressive. 
As for your original intention, what was it?

zelig: Not sure the semantics of topic or pubsub changes here, I thinkk the various bindings merely link the updates on a topic differently to each other as well as allow for multiple sources

zelig:
Revision 3 pushed (22e8325):

- **API section specified** — was a placeholder pointing at bee ethersphere/bee#5435. The gap surfaced while drafting SWIP-61 (#105). Shape follows #5435 generalised from its hardcoded GSOC-ephemeral mode: `GET /pubsub/{topic}` WS upgrade, roles by parameter presence (cohort params ⇒ opener sends `Open`; `owner` ⇒ publisher; bare ⇒ `Subscribe`, spec learned from the `Ack` echo). Serialization adopts the SOC-subscription-family conventions — `swarm-soc-fields` / `swarm-cache-wrapped-chunk` per ethersphere/bee#5497 — making `/mic`//`/moc` (ethersphere/bee#5486) the storage-fed counterparts of the `OWNER`/`SOC_ID` bindings. Key-holding rule: signing is client-side only, the node never holds publisher keys. New conformance item 5: the bridge round-trips both worked configurations.
- **`po_min` is out of `CohortSpec`** — now protocol constant `PO_MIN = 16` (per the thread above: Copilot's proto3 unset-equals-0 footgun, and no use case varies it). Field 7 reserved.

zelig:
Amendment pushed (98e8918), companion to SWIP-65 (#106):

- **ANCHOR dedup soundness**: wrapped-CAC dedup is the guard against unsolicited republication of old SOCs, and is sound only under an application-level requirement — payloads distinct, i.e. the application includes some index in the payload. Without self-indexing the sequence requirement doesn't disappear; it moves above the protocol, unspecified.
- The two "SWIP-65 (forthcoming)" anchors now link #106.

acud:
comments courtesy of claude regarding the part of the implementation that is already on my branch
[2026-08-11-swip-60-feedback.md](https://github.com/user-attachments/files/31156974/2026-08-11-swip-60-feedback.md)

zelig:
Both wire findings taken, and the second one pushed the design further than the fix —
the whole publisher/audience model was rebuilt around it. Rev 4 pushed.

## 1. `Open` / `Subscribe` indistinguishable — taken, envelope

Exactly right, and the silent-success part is the sharp end of it: a receiver that guesses
wrong doesn't fail, it misreads the frame and then answers with a `Status` describing a
different problem. Adopted the envelope your prototype uses:

```proto
message Hello {
  oneof handshake {
    Open      open      = 1;
    Subscribe subscribe = 2;
  }
}
```

Two protocol ids would also have worked, but splitting one-stream-per-(peer, topic) across
two stream names costs more than the `oneof` does. Noted in the framing section so the next
reader doesn't rediscover it.

This also closes your earlier point on the old `Connect`: the establishment message is now
explicitly its own thing, and a subscriber sends nothing but a topic.

## 2. `PublisherAuth` unauthenticated — taken, and `closed` is gone

Your finding was right and the diagnosis was better than the fix would have been. `closed`
was never meant as confidentiality — it was meant to say "every member publishes, there is
no audience" — but the spec's wording *did* promise access control, and the frame couldn't
deliver it. Rather than qualify the sentence, the model changed:

**`Auth` now carries a signature and no address.** The owner is the ecrecover output, so
presenting it is possession of a key rather than a claim about one — identity and proof in
one operation, no challenge round trip:

```
owner = ecrecover( H("bps-join:v1" ‖ topic ‖ admin), signature )
```

Two things worth flagging in that preimage:

- **No libp2p peer id in it, deliberately.** Binding to the node would make the credential
  unreplayable, but it would weld the publishing identity to the node holding the stream —
  the same key unusable from a second node without re-signing, and every join publishing an
  eth-identity/peer-id link. A static preimage is replayable instead, which costs nothing:
  a replayed role is worthless without the signing key, since every frame is still validated
  at `Publish`. Authorship rests on the message signature, never on the handshake.
- **`"bps-join:v1"` is load-bearing.** The same secp256k1 keys sign SOCs over
  `id ‖ wrappedAddress`; without separation a join signature could be reinterpreted as a
  chunk signature, or the reverse.

**`closed` is removed** and replaced by `spectators`, which is enforceable precisely because
`Auth` is recovered rather than asserted. It bounds attendance at a broker and nothing more:
BPS still offers **no confidentiality at any layer** — the broker holds plaintext — and that
is now said plainly in a new Security considerations section instead of being implied away.
A jam is private because it encrypts, not because it refuses spectators.

## 3. The smaller notes — all three taken

- **Dedup horizon**: implementation-defined but MUST be bounded, with the consequence stated
  — a legitimate publisher can overrun the window and replay an evicted message. Applications
  that can't tolerate that carry their own sequencing, which SWIP-65 gives for free.
- **Cohort lifetime**: broker-side, not tied to the opener, reclaimable when unattached. Plus
  a deliberate ending — see `END_OF_STREAM` below, which separates "over" from "the broker
  stopped relaying".
- **Cohort count**: a conformant broker now bounds how many cohorts it will create, not only
  the streams within one. `Open` was indeed an unbounded allocation primitive.

## What else changed, since it touches your branch

The publisher roster **left the cohort spec**. It is dynamic — an admin grants and revokes
while a cohort runs — and the spec is immutable, so it couldn't live there without making
every roster change a new cohort. Grantee identities also aren't public the way an admin's
is: the owner of a stream or a co-edited file is naturally known to subscribers; the other
grantees are not.

It travels instead as admin-signed service messages on a feed the admin owns:

```
owner = admin   id = keccak256("bps-service:v1" ‖ topic ‖ index)

index 0   GENESIS        the CohortSpec, signed by the admin
index n   ROSTER         the full publisher set at version n
last      END_OF_STREAM  the admin closes the cohort, attributably
```

A feed rather than one constant-id slot overwritten in place, because overwriting makes a
stale roster undetectable — which would put forging-by-omission back at the one point that
decides who may write. `Ack` now carries the genesis SOC and the latest service SOC with its
index, so a joiner verifies the cohort *and* the roster against the admin's key, never the
broker's word.

Revocation is two-phase, and the boundary is the moment the reduced roster reaches
subscribers. Before it, the revoked peer has no way to know — its frames are dropped and
**tolerated**, no penalty, connection untouched, because it isn't misbehaving. After it, the
peer has been told on the same feed as everyone else, so publishing is a protocol violation
and the connection is broken. The announcement is what converts an unknowing publisher into a
violating one; disconnecting first would punish a peer for a rule it hadn't been given, and
never publishing the roster leaves the violation unable to begin at all — an ordinary,
visible withholding fault. It also makes the revocation legible to everyone else, who learn
why a publisher fell silent from an admin-signed message rather than from an unattributable
disconnection.

What remains in the spec is immutable policy, and it now names five configurations
explicitly: **jam**, **spectator-jam**, **live-stream**, **group-chat**, **implicit**.

## Back to your review of 4 August — two threads still open

**Topic bindings.** Your point was that pubsub normally gives no semantics to a topic, and
that per-binding validation makes a larger API surface for users. I think rev 4 answers it
better than the argument did: there is now a `MNEMONIC` binding where the topic names the
cohort and constrains nothing — any SOC from any owner. That is the plain pubsub case, it is
the default shape for group chat, and every other binding is opt-in. So the "dumb" semantics
you were asking for is a first-class option rather than something to be recovered by not
using the feature. The other bindings don't change what a topic *is*; they say how updates on
one are linked and whether more than one source can write to it.

**"Remove multi-publisher from this iteration."** I argued in August that a single-publisher
cohort is the harder one, because it's the one that needs authentication. That's now built —
`Auth` plus the genesis message — so the argument has expired and the scoping question is
open on its merits. Worth deciding against the five configurations rather than against the
old spec: `ADMIN_ONLY` (live-stream) and `implicit` need no roster and no service feed at all,
so a first iteration could be exactly those two, with `GRANTED`/`ALL` and the service feed as
the second. Say if that split matches what's on your branch and I'll mark the spec that way.

## What this means for the `bps` branch

Your transport layer survives intact: the `Hello` envelope (identical `oneof`), `Soc` /
`Publish` / `Broadcast`, `Status`, self-contained full-SOC frames, `po_min` as a constant with
field 7 reserved, and your 12 August relaxation of the `ANCHOR` address check under explicit
regimes. `frame.go` and the SOC path shouldn't move at all.

What changes is the policy layer: `PublisherRegime` values, `publisher_list` (field 6, now
reserved), `closed` (field 8, now reserved — `spectators` is field 9), `PublisherAuth` →
`Auth` with a recovered signature, `Ack` carrying the two service SOCs, and the service feed
itself, which is entirely new surface. So `binding.go`, `cohort.go`, `publisher.go` and
`session.go` take most of it, and `bridge.go` needs the service endpoint plus client-side
signing.

One deliberate courtesy: `MNEMONIC` is **appended as `TopicBinding = 5`, not inserted at 1**,
so `ANCHOR`/`SOC_ID`/`OWNER`/`FEED_TOPIC` keep the numbering you already implement. The proto
says so in a comment, in case a later editor is tempted to tidy it.

If you'd rather cut the surface for a first iteration, the split falls out cleanly:
`ADMIN_ONLY` and `implicit` need no roster and no service feed at all, which is close to what
you already have. `GRANTED` / `ALL` and the service feed would then be the second pass.

---

Pushed as `10df5e9`.

zelig:
@acud one implementation note that came out of revising SWIP-61 alongside this — relevant if `Ack` grows in the prototype.

**`Ack` fields 3, 4 and 5 are taken** by rev 4: `genesis`, `service` and `index` (the admin-signed service SOCs a joiner verifies the cohort and roster against). SWIP-61's `candidates` field, which an earlier draft of it numbered 3, is now **field 6** for exactly that reason — [PR #105](https://github.com/ethersphere/SWIPs/pull/105), rev 2 just pushed.

The other thing worth knowing before you touch the multihop side: SWIP-61 no longer keeps publishers at the root. `Publish` travels rootward now, so the depth-1 attachment your prototype inherits from SWIP-60 is a property of singlehop rather than a rule to preserve. Nothing in the current branch needs to change for that today — it only matters once relaying lands.

zelig:
@acud — separate ask: **would you open a draft PR for the `bps` branch?**

I went through it properly and it deserves a review surface. Right now it exists only as a branch, so there is nowhere to comment and nobody but me can see that the spec has a working implementation behind it — which is the strongest argument this SWIP has.

Things I wanted to say inline and couldn't:

- `TestHostileBrokerCannotForge`, `TestOpenRefusesTamperedSpecEcho`, `TestPublisherAuthIsNotACredential` — testing the security *properties* adversarially rather than the happy path. That last one arrives at the same framing rev 4 now writes down: the handshake declaration is an early refusal, and the real gate is the recovered owner at `Publish`. Good to see it reached from the implementation side independently.
- The `serve` teardown reasoning — `writeCtx` bridging both quit channels so a peer that stops draining its flow-control window cannot park the goroutine, and `Reset` rather than `FullClose` because bee's `FullClose` reads and would give you two concurrent readers. Explained in place, which is what makes it reviewable.
- Declining to label metrics by peer address, so a remote peer cannot mint unbounded time series. Right call, and the reason is worth keeping in the comment.
- `SpecEqual` comparing the publisher list as a **set**, for two clients assembling the same invite in different orders. The spec never said that and should have.

Two things a review would actually be about:

1. **Blocklisting.** SWIP-60 says repeated invalid ⇒ disconnect; the branch drops and logs, and your comment notes the policy wants per-peer state in memory. Fair, but it is a conformance gap worth tracking rather than leaving in a comment.
2. **A spec question your code exposes.** Under an explicit regime `anchorBinding.qualifies` returns `nil` unconditionally — the topic becomes a mere rendezvous. With rev 4's new `MNEMONIC` binding, `ANCHOR` + roster and `MNEMONIC` + roster now differ *only* in the dedup rule: wrapped CAC versus chunk address. That may be exactly the right distinction, or `ANCHOR`-under-explicit may now be redundant. I would rather decide that with you than assert it in the spec.

And it is the natural place to settle the scoping question from my previous comment — which configurations land first. `ADMIN_ONLY` and `implicit` need no roster and no service feed at all, which is close to what you already have.

SWIPs/swip-60.md
```diff
@@ -0,0 +1,700 @@
...
+created: 2026-08-03
+---
+
+<!-- Base SWIP of the Broadcast Pub/Sub (BPS) family: the decomposition of the monolithic
+PubSub SWIP (ethersphere/SWIPs PR #93) into work-package-sized SWIPs. Companion protobuf:
+assets/swip-60/bps.proto. -->
+
+- **Business line**: real-time topic streams for dApps without storing chunks or polling —
+  enough on its own for the five cohort shapes it defines: **jam** (a closed set of authors:
+  collaborative remix editing, a strudel livecoding session, a multiparty game),
```
nugaon: > strudel livecoding session

?? strudel is a food or not? sorry I haven't heard this combination of the words before...

nugaon: instead of `multiparty game` the `multiplayer game` would be better.

SWIPs/swip-60.md
```diff
@@ -0,0 +1,700 @@
...
+the moment the first peer contacts a BPS-speaking full node with a topic, and **immutable for
+the cohort's lifetime**. There is no mode enum; **modes are combinations of these
+parameters**.
+
+| parameter | values | meaning |
+|---|---|---|
+| `topic` | 32 bytes | interpreted per `binding` |
+| `binding` | `MNEMONIC` / `ANCHOR` / `SOC_ID` / `OWNER` / `FEED_TOPIC` | what the topic binds to; fixes which SOCs qualify as messages and the dedup rule |
+| `admin` | eth address | the opener, the cohort's authority, and a member of its publisher set. **Absent ⇒ implicit authorship**, and the two fields below do not apply |
+| `publishers` | `ADMIN_ONLY` / `GRANTED` / `ALL` | who may author besides the admin |
```
nugaon: the topic always should be derived from the SOC address or tied to that - at least this is the original idea. any other types that requires authentication state handling can came later IMO.

SWIPs/swip-60.md
```diff
@@ -0,0 +1,700 @@
...
+| **implicit** | absent | — | true | whoever the binding's SOC shape admits |
+
+`spectators` does real work only in the `GRANTED` and `ADMIN_ONLY` rows — which is exactly
+the audience / no-audience distinction. Under `ALL` and under implicit authorship every
+attached peer is already a potential author, so excluding non-publishers excludes nobody;
+openers MUST set it true there.
+
+**The admin is always in the publisher set**, and being a publisher obliges nobody to
+publish — no peer waits on another — so a practically non-publishing **moderator** needs no
+role of its own: it is simply an admin that never sends.
```
nugaon: why not send anything and who is an admin?

SWIPs/swip-60.md
```diff
@@ -0,0 +1,700 @@
...
+  from any owner qualifies (dedup on chunk address). This is what `ALL` needs. Authorship is
+  unrestricted but never *unattributable*: every message is still SOC-signed, so a group chat
+  knows exactly who said what without there being an authorised set to check it against.
+- **`ANCHOR`** — topic = full SOC/GSOC address; all messages share one address (dedup on
+  the wrapped CAC — the guard against unsolicited republication of old SOCs, sound only
+  under an application-level requirement: payloads are distinct, i.e. the application
+  includes some index in the payload).
+- **`SOC_ID`** — topic = SOC id; any owner with `PO(socAddr(id, owner), anchor) ≥ PO_MIN`
+  qualifies (dedup on chunk address).
+- **`OWNER`** — topic = `keccak256(owner)`; any id under the same PO constraint — MIC
```
nugaon: IMO these should be named `GSOC`, `MOC` and `MIC`, respectively. 
It would simplify with the already used terms.

SWIPs/swip-60.md
```diff
@@ -0,0 +1,700 @@
...
+| `SOC_ID` | MOC | **many** | any owner that mines `PO(socAddr(id, owner), anchor) ≥ PO_MIN`; the id is fixed, the owner varies |
+
+Where authorship is explicit and dedup is on the wrapped CAC (`ANCHOR`), the SOC id does no
+protocol work: it is **unconstrained**, and publishers MAY use it as a plain sequence number.
+The full sequential construction — signed as a feed update, carried as a bare index, making
+missed updates detectable and recoverable — is **self-indexed feeds,
+[SWIP-65](https://github.com/ethersphere/SWIPs/pull/106)**.
+
+The proximity constraint for implicit bindings is a **protocol constant**, not a cohort
+parameter: `PO_MIN = 16`. (Making it a parameter invited proto3's unset-equals-0
```
nugaon: it can be greater than 16 as well

nugaon (commented):
I would break this SWIP into smaller pieces as the [POC](https://github.com/ethersphere/bee/pull/5435) also just implemented the GSOC messaging only, that we need in most of the use-cases ATM.

zelig:
**Rev 5 — SWIP-60 now extends SWIP-74 (BPS-lite, #111) and changes nothing in it.** The proto is revision 8, derived from SWIP-74's block.

- **One handshake frame, `Join`**, carrying the full `CohortSpec`; `Hello`, `Open` and `Subscribe` are gone. Cohorts are keyed by the whole spec: a spec no live cohort has creates one, a byte-identical spec attaches, and there is no "unknown topic". Pre-creating a topic under a wrong admin squats nothing.
- **`Ack` is a status.** No spec echo — the joiner brought the spec. The latest service SOC is delivered as the first `Message` on every newly attached stream instead, so a joiner still learns the roster from the admin, not the broker *(open: this is the one new rule; the alternative is to keep it in `Ack`'s reserved fields)*.
- **`GENESIS` is gone**: with the spec in every `Join` it had nothing left to prove. The service feed starts at index 0 with the first `ROSTER` or `END_OF_STREAM`, and each service message carries its index in the payload.
- **The publisher regime shrinks to one value, `ALL`.** Unset, with an admin: the admin publishes, and whoever its roster ever names — a cohort is multi-publisher iff a roster is ever published, and nobody needs to know in advance. `ADMIN_ONLY` and `GRANTED` go; live-stream and spectator-jam are one spec.
- **`spectators` reverts to `closed`** (field 8, unset = open audience). As a proto3 bool whose unset value is false, `spectators` read every SWIP-74 spec — which never sets it — as a closed cohort.
- **The chunk travels as opaque chunk data**: one `Message{soc}` both directions replaces `Publish`, `Broadcast` and the field-level `Soc`. Deliveries go to every stream except those bound to the publishing identity.
- **Auth is bound to the stream, not the connection**; a spectator carrying an identity is promoted in place when a later roster names it.
- Lifetime and capacity bounds (per cohort, per broker, per peer connection, inactivity deadline) are SWIP-74's, referenced rather than restated; the admin's `Join` is admitted past the per-cohort bound. The feed **cursor stays SWIP-74's stricter special case**: a full broker dedups on chunk address (several publisher feeds on one topic, SWIP-61 reordering) and passes retransmits through — duplicates are counted as retransmits, never as invalid.
- Service SOCs ride the `Message` frame and are recognised by id before the content path; under `FEED_TOPIC` a feed update's id slot is a bare index, a service SOC's is its full id. `ANCHOR` under explicit authorship performs no address check (topic = rendezvous), stated explicitly now.
- API: the spec parameters are required on every session (the invite carries them); worked URLs updated; `closed` replaces `spectators`.

Conformance items 1, 2, 5, 6, 7, 8 updated accordingly; SWIP-61 is to be re-based on the `Message` frame (a rename of `Publish`/`Broadcast`). SWIP-65's carriage rule still needs its one-line exception for the cursor (separate PR).

zelig:
**Rev 6 — the claim handshake, from SWIP-74 rev 3; proto revision 9.**

- `Auth`-in-`Join` is gone. A publisher declares its address in `Join.addr`, receives a broker-derived challenge in `Ack`, and claims by signing it together with the broker's overlay and its cursor (`Claim{addr, index, auth}`); a returning publisher claims in the `Join` itself. No reply.
- "The first frame settles the cohort; the claim settles the role": the role table is now by claim outcome. `closed` no longer refuses at `Join`: a joiner is silent — receives nothing, not even the roster — until its claim recovers to a rostered address, and is disconnected otherwise.
- Promotion is by an explicit claim: a peer named in a roster claims on the stream it holds, or in its next `Join`. Nothing is bound ahead of time.
- Under `ALL` there is no claim: a stream declares the address it publishes as, and every message is validated against it by the SOC's hash and signature. A replayed group-chat join buys entry, which is open, and not one message under the borrowed name.
- Feed publishers under explicit authorship have SWIP-74's cursor, one per publisher feed, set forward by the claim; the other bindings keep the bounded dedup window.
- One extra stream over the fan-out bound per absent legitimate publisher (admin or rostered), with a claim deadline.
- API: `auth` becomes `addr`; the node hands the session the challenge and its broker's overlay, the dApp signs, the node sends the claim (marked (?) — the bridge needs a control frame for this). Worked URLs updated.
- Security rewritten: what the claim protects (identity, not history), the case-by-case table lives in SWIP-74, the transport precondition (record must name the authenticated peer ID) is normative.
- Proto rev 9: `Auth{r, s, v}`, `Claim`, `Join{cohort, addr, claim}`, `Ack{status, challenge}`, `Message{address, data}`; `CohortSpec` renumbered to follow SWIP-74 (`admin = 3`, then `publishers`, `history`, `closed`); no reserved statements.
- Conformance 6–8 updated; SWIP-61 to be re-based on `Message` and the claim (one paragraph added there).

PR #105, opened 2026-08-05
SWIP-061: BPS multihop — FCFS multicast tree

zelig:
Extends SWIP-60 (#104): removes the singlehop audience ceiling. A full node at capacity answers `Subscribe` by probing its own subtree over existing streams and returning the two shallowest attachment points — the joiner attaches to **both**, so every node keeps two parents (typically sisters): churn is masked, not suffered, self-healing is re-running the join, and withholding shows by comparing feeds. Join is ideally a two-step process regardless of tree size.

**Key properties**: forwarders are subscribers (accepting your first child = becoming a relay); no standing control state — idle cohorts are silent, probes carry depth in the frame itself; no depth bound (per-hop pricing steers joins rootward; closed cohorts are structurally singlehop); `CohortSpec` untouched; three control frames (`Reparent`, `Probe`, `Candidates`) fill the reserved `Broadcast` range; single new `Ack` field. Unmodified SWIP-60 clients attach as conformant (single-parented) leaves.

**Deliberately out of scope** (own SWIPs): bandwidth incentives, message history (pre-subscription delivery), broker discovery (SWIP-59, #103), NAT traversal (transport: circuit relay + DCUtR).

Wire: `assets/swip-61/bps.proto` — the SWIP-60 proto with the multihop control plane filled in. Implementation groundwork: bee ethersphere/bee#5435.

### DIFF
--- a/SWIPs/assets/swip-61/bps.proto
+++ b/SWIPs/assets/swip-61/bps.proto
@@ -34,8 +34,8 @@ option go_package = "github.com/ethersphere/bee/v2/pkg/bps/pb";
 enum TopicBinding {
   TOPIC_BINDING_UNSPECIFIED = 0; // invalid on the wire (see header note)
   ANCHOR     = 1; // topic = full SOC/GSOC address; dedup on the wrapped CAC
-  SOC_ID     = 2; // topic = SOC id; any owner with PO(addr, anchor) >= po_min
-  OWNER      = 3; // topic = SOC owner; any id with PO(addr, anchor) >= po_min (MIC)
+  SOC_ID     = 2; // topic = SOC id; any owner with PO(addr, anchor) >= PO_MIN
+  OWNER      = 3; // topic = SOC owner; any id with PO(addr, anchor) >= PO_MIN (MIC)
   FEED_TOPIC = 4; // id = keccak256(topic ‖ index); graffiti MIC / feed streams
 }
 
@@ -53,6 +53,8 @@ enum PublisherRegime {
 // NOTE: broker capacity is NOT a cohort parameter — a cohort cannot dictate a
 // remote node's connection count. Each broker enforces its own per-topic
 // stream limit and answers FULL when it is exhausted.
+// NOTE: the proximity constraint for implicit bindings is a protocol
+// constant, PO_MIN = 16 — not a cohort parameter.
 message CohortSpec {
   bytes           topic          = 1; // 32 bytes, meaning per binding
   TopicBinding    binding        = 2;
@@ -61,7 +63,7 @@ message CohortSpec {
   bytes           admin          = 5; // 20-byte eth address; set iff EXPLICIT_*
   repeated bytes  publisher_list = 6; // 20-byte eth addresses, excl. admin;
                                       // set iff EXPLICIT_LIST
-  uint32          po_min         = 7; // proximity order for implicit bindings (default 16)
+  reserved 7;                         // was po_min — now protocol constant PO_MIN = 16
   bool            closed         = 8; // no audience: subscribers restricted to the publishers
 }
 
--- a/SWIPs/assets/swip-61/bps.proto
+++ b/SWIPs/assets/swip-61/bps.proto
@@ -1,0 +1,173 @@
+// Broadcast Pub/Sub (BPS) — protocol messages and types.
+// Spec: swip-bps-singlehop.md (SWIP-60, base protocol) and
+// swip-bps-multihop.md (SWIP-61, multihop control plane).
+//
+// Revision 2 (2026-08-05), after review on PR #104: Connect split into
+// Open/Subscribe (subscribers carry no cohort metadata), broker capacity
+// removed from CohortSpec (it is broker-side policy, not a cohort parameter),
+// Ping dropped (liveness/RTT are transport concerns), and every frame carries
+// the full SOC (no handshake/data split). Field numbers renumbered — the
+// draft has no deployed compatibility surface.
+//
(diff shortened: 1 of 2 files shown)


zelig:
Amendment pushed (1ca57ae), companion to SWIP-65 (#106):

- **Dual parenting is not required on self-indexed cohorts**: a gap there is detectable at the next frame and recoverable from storage, so the contract is satisfied by deliver-or-recover. The second parent becomes a per-subscriber latency/cost choice — single-parented nodes were already well-formed (SWIP-60 leaves), and a single-parented relay endangers only itself. Conformance items 5–6 bind only nodes that maintain one. Dual parenting remains the costlier-but-faster product for the gap-intolerant live edge.

zelig:
Rev 2 pushed (`00aa789`), realigning this SWIP with [SWIP-60 rev 4](https://github.com/ethersphere/SWIPs/pull/104) and lifting the invariant that confined publishers to the root.

## Two amendments to SWIP-60 now, not one

The capacity rule is unchanged. The new one: **`Publish` travels rootward**, forwarded hop by hop to the broker, which validates, dedups and broadcasts it back down.

SWIP-60's direct attachment of publishers was a consequence of depth = 1, not an invariant — so multihop now restructures **both** directions rather than the audience side only. That is what an everyone-publishes cohort needs in order to exist at all: SWIP-60's `ALL` (group chat), and any `GRANTED` cohort larger than one node's capacity, were previously capped by the broker's stream limit on the publisher side no matter how far the tree grew.

`Open` still goes to the broker — opening a cohort *is* contacting the root — but its admin is thereafter an ordinary publisher and may sit at any depth.

A new section, **Publishing from depth**, settles three things:

- **Validation is authoritative at the root**, because dedup is a statement about what the cohort has *seen*, and only the root sees everything: a relay applying the rule locally would suppress a message some sibling subtree never received. Relays MAY and SHOULD run the *stateless* checks early — signature, binding, authorship — since carrying a doomed frame up d hops wastes every one of them. Stateless checks anywhere, stateful ones where the state is. A relay that forwards everything is still conformant; one that drops a valid frame is withholding.
- **Publish up both parents.** The upstream twin of the two downstream feeds: a single relay cannot silently swallow a publish, and the dedup rule absorbs the double at the sisters' shared parent.
- **Depth is priced, not bounded** — see below.

**Revocation across a tree**: the roster update is an ordinary broadcast, so enforcement lands at the revoked publisher's *parents* — they are the nodes accepting its `Publish` frames — and the revokee learns of its own revocation from the same broadcast at the same time. That is what makes SWIP-60's two-phase rule fair at any depth.

## The "why no depth bound" argument, re-anchored

It used to rest on the closed cohort being *structurally singlehop*. That anchor is gone twice over: `closed` no longer exists in SWIP-60, and publishers no longer attach to the root. **No configuration requires depth 1.**

The argument now stands on the price alone, and is stronger for it — with `Publish` travelling rootward, depth-scaled upstream pricing bills a *publisher* for its depth, which is the case where the incentive bites hardest.

## What a relay owes a joiner

SWIP-60 rev 4's `Ack` carries the echoed `CohortSpec`, the admin-signed genesis SOC and the latest service SOC with its index. A relay MUST pass on all three unchanged — it holds them from its own join, and keeps the service SOC current from broadcasts it already receives.

**Depth must cost a subscriber nothing in what it can verify.** A leaf ten hops out checks the cohort and the roster against the admin's key exactly as the broker's own child does, so extra intermediaries add withholding points and no forging points. A relay serving a stale service SOC is withholding, which its subscriber detects as an index gap. `spectators: false` cohorts work at depth for the same reason: any node answering a `Subscribe` holds the roster.

## Wire — one collision caught

`Ack.candidates` is now **field 6, not 3**. SWIP-60 rev 4 uses fields 3, 4 and 5 for `genesis`, `service` and `index`, so the number this draft previously claimed would have collided with them.

`assets/swip-61/bps.proto` is now regenerated from the SWIP-60 rev 4 file rather than patched, so the two stop drifting.

## Review items cleared

Thanks @Copilot — all taken:

- front matter split into `type: Standards Track` + `category: Networking`, matching swip-19/20
- proto header pointed at `swip-60.md` / `swip-61.md` instead of the nonexistent `swip-bps-singlehop.md` / `swip-bps-multihop.md`
- establishment comments corrected: **any** cohort node answers `Subscribe`, not only the broker
- `DCUtR` casing normalised throughout

The `**(?)**` on graceful leave is resolved rather than removed: a departing relay reparents each child to **its own parents** — known-live, known-compatible, and one level shallower, so the subtree gets shallower rather than deeper. Handing children to one another would deepen the tree and has no answer for the first child.

Both mermaid diagrams are verified to parse under mermaid 11. Worth noting for anyone editing SWIP diagrams: a **semicolon in note text silently breaks mermaid** — it is a statement separator, so `Note over A,B: one; two` is a parse error and GitHub falls back to showing the raw source. That defect had been in SWIP-60's diagram since it was written; it is fixed there now.

zelig:
One paragraph added under *Publishing from depth* — **Claims travel rootward too** — to keep SWIP-61 consistent with the claim handshake now in SWIP-74 rev 3 and SWIP-60 rev 6: a publisher proves its key to the node it attaches to (that node's own challenge, signed with that node's overlay and the publisher's cursor), and the attachment node forwards the claim rootward iff the address is the admin's or in the roster it holds, so that a publication travels rootward only on streams claimed for its owner, hop by hop. Under `ALL` nothing is forwarded: no claim exists, every node authenticates each publication against the address its child stream declared. The full re-base of this SWIP onto the `Message` frame (SWIP-60 rev 5 folded `Publish`/`Broadcast` into one) is still pending and marked (?).

